PR #191, opened 2015-01-08
feat(api): Add `action=force_auth` to GET /v1/authorization.

### LINKED ISSUE #190: Support "force auth" in GET /v1/authorization
We need the ability to force the user to sign in as a specific user. This option is only supported for the `signin` action. If the force auth option is specified, the user should be redirected to `/force_auth` on the content server. The content server should prevent the user from editing the email address, and should not allow the user to sign up if the email is not registered.

For more background, see https://github.com/mozilla/fxa-content-server/issues/1957, https://github.com/mozilla/fxa-oauth-server/issues/181, and https://bugzilla.mozilla.org/show_bug.cgi?id=1104304.

Two proposals for the relier to indicate they wish to use the force auth flow with the `signin` action:

1. `force_email=<xxx>`
1. `email=<xxx>&force_auth=true`

shane-tomlinson:
`action=force_auth` redirects to the `/force_auth` page on the content server. On the content server, the user must sign in with the email specified by the `email` query parameter.

fixes #190 

See mozilla/fxa-content-server#1957

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -260,8 +260,9 @@ content-server page.
 - `state`: A value that will be returned to the client as-is upon redirection, so that clients can verify the redirect is authentic.
 - `redirect_uri`: Optional. If supplied, a string URL of where to redirect afterwards. Must match URL from registration.
 - `scope`: Optional. A space-separated list of scopes that the user has authorized. This could be pruned by the user at the confirmation dialog.
-- `action`: Optional. If provided, should be either `signup` or `signin`. Send to improve user experience, based on whether they clicked on a Sign In or Sign Up button.
-- `email`: Optional. If provided, this email address will be pre-filled into the account form, but the user is free to change it.
+- `action`: Optional. If provided, should be `signup`, `signin`, or
+  `force_auth`. Send to improve user experience, based on whether they clicked on a Sign In or Sign Up button. `force_auth` is used to force the user specified by `email` to sign in. If `force_auth` is specified, the user is unable to modify the email address, and is unable to sign up if the user does not exist.
+- `email`: Optional if `action` is `signup` or `signin`. Required if `action` is `force_auth`. If provided, this email address will be pre-filled into the account form, but the user is free to change it.
 
 **Example:**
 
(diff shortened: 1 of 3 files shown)


rfk:
Rebasing on latest master should fix the unrelated travis failure, thanks to #193

docs/api.md
```diff
@@ -260,8 +260,9 @@ content-server page.
 - `state`: A value that will be returned to the client as-is upon redirection, so that clients can verify the redirect is authentic.
 - `redirect_uri`: Optional. If supplied, a string URL of where to redirect afterwards. Must match URL from registration.
 - `scope`: Optional. A space-separated list of scopes that the user has authorized. This could be pruned by the user at the confirmation dialog.
-- `action`: Optional. If provided, should be either `signup` or `signin`. Send to improve user experience, based on whether they clicked on a Sign In or Sign Up button.
-- `email`: Optional. If provided, this email address will be pre-filled into the account form, but the user is free to change it.
+- `action`: Optional. If provided, should be `signup`, `signin`, or
+  `force_auth`. Send to improve user experience, based on whether they clicked on a Sign In or Sign Up button. `force_auth` is used to force the user specified by `email` to sign in. If `force_auth` is specified, the user is unable to modify the email address, and is unable to sign up if the user does not exist.
```
rfk: If we add any more options here, we should consider just linking to their official documentation on the content-server side.

lib/routes/redirect.js
```diff
@@ -6,11 +6,21 @@ const url = require('url');
...
 const config = require('../config');
 
+function actionToPathname(action) {
+  if (action === 'signup') {
+    return 'signup';
+  } else if (action === 'force_auth') {
+    return 'force_auth';
+  }
+
+  return 'signin';
```
rfk: Should we consider erroring out if an invalid `action` is specified?  I can imagine people accidentally typoing this as e.g. `force_signin`.

docs/api.md
ckarlof: This is the official documentation. :) The OAuth API is endpoint that reliers integrate with.

lib/routes/redirect.js
ckarlof: Yes, I like fast failing on invalid actions.

rfk:
r+ assuming travis goes green after a rebase.  Feel free to treat my nits as follow-up bugs rather than fixing inline.

PR #193, opened 2015-01-08
chore(dependencies): bump hapi version

vladikoff:


pdehaan:
That's a lot of deletions. Did you run shrinkwrap with the `--dev` flag? (or maybe Hapi just removed some dependencies)

vladikoff:
>  Did you run shrinkwrap with the --dev flag?

Ah no, updated with that now.

pdehaan:
I should probably just get into the habit of adding a "shrinkwrap" script to package.json, so we only need to remember to do `npm run shrinkwrap` in each project and it will be consistent and we wouldnt need to remember to use the various flags.

rfk:
thanks @vladikoff!

vladikoff:
Thanks for merging!

PR #198, opened 2015-01-21
feat(api): allow destroying token without client_secret

seanmonstar:
A problem came up that I'm trying to revoke tokens created by implicit grants, so I don't have the client secret available. I looked at Google's API, and they don't require any authorization either (if you already have a token, things gon burn): https://developers.google.com/accounts/docs/OAuth2WebServer#tokenrevoke

@rfk r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -381,7 +381,6 @@ destroy the token afterwards. A client can use this route to do so.
 #### Request Parameters
 
 - `token` - The hex string token.
-- `client_secret` - The client secret used to get the token originally.
 
 **Example:**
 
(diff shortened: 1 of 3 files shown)


ckarlof:
I like it. @rfk is on PTO until next week. I don't suspect this is blocking anything, but if it is, we find another reviewer this week.

rfk:
:+1:

rfk:
Although...will we have any clients in the wild that are passing the client_secret to this API.  Perhaps we should accept it and silently ignore it (which may be the default behaviour of hapi anyway, but it's worth thinking about explicitly)

seanmonstar:
@rfk I want to say this does what you said, but I don't know for sure: https://github.com/mozilla/fxa-oauth-server/blob/master/lib/server.js#L30

rfk:
Let's add an explicit test to this effect, just to be sure

rfk:
I filed #202 to follow up on this

PR #200, opened 2015-01-23
WIP feat(api): support for client owners

vladikoff:
WIP

- [ ] Limit access to clients that you own.
- [ ] Add MySql support


lib/routes/client/register.js
```diff
@@ -41,6 +42,8 @@ module.exports = {
   handler: function registerEndpoint(req, reply) {
     var payload = req.payload;
     var secret = unique.secret();
+    var authHeader = req.headers.authorization;
```
seanmonstar: You can access this via hapi's `auth` property. `req.auth.credentials` will be an object like this: https://github.com/mozilla/fxa-oauth-server/blob/master/lib/token.js#L18

seanmonstar:
@vladikoff any help or guidance I can give here?

vladikoff:
going to continue with this in https://github.com/mozilla/fxa-oauth-server/pull/222

PR #201, opened 2015-01-24
fix(api): reject requests with bad content-types

### LINKED ISSUE #199: request sent with "text/plain" content-type causes 500 error
I was triggering 500 errors locally when I hit the `POST /v1/authorization` endpoint with an incorrect `text/plain` content type header.

seanmonstar:
Only blank and application/json are allow.

Closes #199 

r? @zaach


zaach:
:+1:

Issue #183, opened 2014-12-10
DB migration infrastucture


rfk:
Let's get some DB migration infrastructure in place for this repo, we've now got at least two open issues that could make use of it.  We could consider just copying what's done in https://github.com/mozilla/fxa-auth-db-server as the default option.

seanmonstar:
I'm find with the way auth-db does it. @ckolos any opinion, since you can have input *before* we make something crazy?

PR #205, opened 2015-01-27
feat(api): add `last_auth_at` to token response schema.

### LINKED ISSUE #181: Provide a mechanism for reliers to learn when the user last authenticated with the auth server
Some reliers need to know at what time the user last successfully entered her password. E.g., if a user on Marketplace enters her purchase PIN incorrectly too many times in a row, they force her to re-enter her password to reset the PIN. This "last authentication time" needs to be communicated to Marketplace in some manner (https://bugzilla.mozilla.org/show_bug.cgi?id=1104304). 

Luckily, the FxA auth server includes this time in signed certs (https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-12). 

It's not clear what is the best way to communicate it to Marketplace, but a simple one would be to return the `authAt` value in the [the response from the OAuth server when the relier exchanges a code for a token](https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1token). 

**Include the time in the response when  https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1token

rfk:
The `last_auth_at` field tells the relier the freshness of the fxa-auth-server
session that was used to complete the flow.  This can be used in conjunction
with `action=force_auth` to ensure that the user has explicitly re-entered
their password during the oauth flow.

fixes #181


rfk:
This is a proof-of-concept for #181.

It is *not* ready for merge, because it requires a db schema change and we don't have the infra set up for that yet; see #183.  I'm making a PR for discussion and for testing the MySQL backend in travis.

The overall flow that this is part of is described in https://bugzilla.mozilla.org/show_bug.cgi?id=1104304#c7, but basically the relier can force the user to re-enter their password by:

  * including `action=force_auth` when initiating the oauth dance
  * checking that `last_auth_at` is recent enough when concluding the oauth dance

Note that the `last_auth_at` value is transient and token-specific; it's not a property of the token so much as it's a property of the particular oauth dance that was completed to generate it.

@seanmonstar and @shane-tomlinson  thoughts?

shane-tomlinson:
Is `fxa-lastAuthAt` client generated or server generated? My guess is server generated, but it is hard for me to tell tracing through the auth server code. If client generated, we will likely have to handle clock skew (user's clocks are all over the place).

rfk:
`fxa-lastAuthAt` is generated by the fxa-auth-server - it's basically the creation timestamp of the sessionToken used to generate the assertion.

If it were client generated, we'd have bigger problems than clock skew - it would defeat the security we're trying to build here, since clients could forge to to avoid re-entering the password :-)

rfk:
note to self: need to sanity-check size of codes table in prod, to see how big a deal this ALTER TABLE is going to be

docs/api.md
```diff
@@ -362,6 +362,7 @@ A valid request will return a JSON response with these properties:
 - `access_token`: A string that can be used for authorized requests to service providers.
 - `scope`: A string of space-separated permissions that this token has. May differ from requested scopes, since user can deny permissions.
 - `token_type`: A string representing the token type. Currently will always be "bearer".
+- `last_auth_at`: An integer giving the time at which the user last authenticated to the Firefox Accounts server, as seconds since epoch.
```
seanmonstar: Seconds, not milliseconds? I ask because JavaScript likes to output timestamps in millis, eg `Date.now()`. Either way, mind bolding the unit so people can find it quickly?

lib/db/memory.js
```diff
@@ -135,12 +136,14 @@ MemoryStore.prototype = {
     delete this.clients[unbuf(id)];
     return P.resolve();
   },
-  generateCode: function generateCode(clientId, userId, email, scope) {
+  generateCode: function generateCode(clientId, userId, email, scope,
+                                      lastAuthAt) {
```
seanmonstar: Side note: I hate it when code gets this long. There is **no** good option of formatting, imo.

lib/db/schema.sql
```diff
@@ -18,6 +18,7 @@ CREATE TABLE IF NOT EXISTS codes (
   INDEX codes_user_id(userId),
   email VARCHAR(256) NOT NULL,
   scope VARCHAR(256) NOT NULL,
+  lastAuthAt BIGINT NOT NULL,
```
seanmonstar: `BIGINT` vs `TIMESTAMP` right below it?

lib/db/memory.js
rfk: aye; I considered passing in a "claims" object with both email and lastAuthAt as properties, but this seemed like a layering violation

docs/api.md
rfk: The auth-server does indeed give it to us in seconds, so I think we should be consistent with that.  Will bold.

lib/db/schema.sql
rfk: I'm hesitant to do any transformation of coercion on this value on the off change that we stuff it up; `BIGINT` is what the auth-server uses for this value so I just copied it across here.

docs/api.md
```diff
@@ -362,6 +362,7 @@ A valid request will return a JSON response with these properties:
 - `access_token`: A string that can be used for authorized requests to service providers.
 - `scope`: A string of space-separated permissions that this token has. May differ from requested scopes, since user can deny permissions.
 - `token_type`: A string representing the token type. Currently will always be "bearer".
+- `last_auth_at`: An integer giving the time at which the user last authenticated to the Firefox Accounts server, as **seconds since epoch**.
```
kumar303: is this a UTC unix timestamp? If not, can you make it one? It would be a lot easier to work with.

lib/db/schema.sql
rfk: Hmm, actually the `BIGINT` is because fxa-auth-server stores a millisecond timestamp, we probably don't need a column that big for just the seconds...

rfk:
> is this a UTC unix timestamp?

It's `(+Date.now()) / 1000` as seen by the fxa-auth-server, so yes, I believe so.  I'll add a clarifying comment in the docs.

kumar303:
ok cool, I think that should be a UTC timestamp.

rfk:
I've rebased this branch atop #208 to test deployment of this as a db migration; the diff view will be a little messy until that gets merged back to master.

rfk:
This is now live on a custom dev server setup at:

  https://lastauthat.dev.lcip.org/
  https://oauth-lastauthat.dev.lcip.org/

(You should be able to intuit any other necessary URLs by reference to https://developer.mozilla.org/en-US/Firefox_Accounts#Latest_development_%28updated_continuously_from_master%29)

It seems to be working for me, correctly reflecting the `last_auth_at` through a token exchange.

rfk:
CldSvcOps bug to sanity-check impact of this ALTER TABLE: https://bugzilla.mozilla.org/show_bug.cgi?id=1128404

rfk:
(added triple-heart label because it would be great to have this on the next train if possible)

ckarlof:
@seanmonstar, please review. This needs to ship in train-30 to support Marketplace Desktop payments, which is cutting tomorrow, 2/2.

ckarlof:
In #181, @seanmonstar  said:

> @ckarlof so, if we record the timestamp when we generate a token to give to Marketplace, that won't necessarily stay updated. They could sign into Markplace on Jan 1, and that's the time we would record. Perhaps Jan 20th, they did an action needing a re-auth, and Marketplace asks us the lastAuthAt value, we'd say Jan 1. Even if the user happily logged in and fiddled with their avatar earlier that day.

I was somewhat deaf to this earlier, but I now see @seanmonstar's point that exposing this value as `lastAuthAt` is not accurate. What we're exposing is the authentication time associated with this token, but not the time which the user last authed at overall. The former has clearer value, but latter seems somewhat worthless if you consider it. :)

Thoughts about a name change? `authed_at`? `auth_time`?

ckarlof:
> This needs to ship in train-30 to support Marketplace Desktop payments, which is cutting tomorrow, 2/2. 

Since OAuth is decoupled from the rest of the stack, we don't absolutely need to get this done today. However, we should get it cut and pushed in the next week or so. In particular, I'd like for us to be happy with the name under which this value is exposed in the OAuth API.

kumar303:
> What we're exposing is the authentication time associated with this token, but not the time which the user last authed at overall

Ah, I didn't know that. This makes more sense security-wise! Could you update the docs to clarify that the timestamp is only associated with the authorization token?

seanmonstar:
> This makes more sense security-wise!

It's not on purpose :). It's a matter of OAuth not knowing about every login the user makes. It only knows when such a login was required to provide an OAuth token somewhere else.

lib/db/mysql/index.js
```diff
@@ -28,67 +31,72 @@ function MysqlStore(options) {
...
-    conn.changeUser({
-      user: options.user,
-      password: options.password,
-      database: database
-    }, function(err) {
-      if (err) {
-        logger.error('changeUser', err);
+    if (patcher.currentPatchLevel !== patch.level) {
+      if (patcher.currentPatchLevel !== patch.level + 1) {
+        err = 'unexpected db patch level: ' + patcher.currentPatchLevel;
```
seanmonstar: since we're rejecting with this `err`, let's make it an `Error` object.

lib/db/mysql/index.js
```diff
@@ -28,67 +31,72 @@ function MysqlStore(options) {
...
-    logger.verbose('changeUser');
-    conn.changeUser({
-      user: options.user,
-      password: options.password,
-      database: database
-    }, function(err) {
-      if (err) {
-        logger.error('changeUser', err);
+    if (patcher.currentPatchLevel !== patch.level) {
+      if (patcher.currentPatchLevel !== patch.level + 1) {
```
seanmonstar: This tripped me for a second. I see now that if our patch level is 1 above the one in the DB, it's no problem because we'll be upgrading. Perhaps a comment to that effect?

lib/db/mysql/index.js
```diff
@@ -28,67 +31,72 @@ function MysqlStore(options) {
...
+  options.createDatabase = options.createSchema;
+  options.dir = path.join(__dirname, 'patches');
+  options.metaTable = 'dbMetadata';
+  options.patchKey = 'schema-patch-level';
+  options.patchLevel = patch.level;
+  options.mysql = mysql;
+  var patcher = new MysqlPatcher(options);
+
+  return P.resolve().then(function() {
+    return P.promisify(patcher.connect, patcher)();
```
seanmonstar: Any reason L88 can't just become L87?

```js
return P.promisify(patcher.connect, patcher)().then(...)
```

seanmonstar:
@ckarlof I thought this through again. I've realized there's a mistake in my concern. The Marketplace will be requesting a new token, forcing auth, to trigger this. So, the recorded value most likely could never be that old. The one case that it could is if a user started this flow, but didn't return immediately to the Marketplace, and possible authed again later with Auth server. The time recorded in the code would older than the newest auth. In practice, this doesn't seem like it would matter, as this value will still tell Marketplace if they authed later than the timestamp of the session they started with Marketplace.

Still, since it's not technically the last time the user authed, I'd rather if we named it `auth_at`. It correctly conveys the time the user had authed with the auth-server when we generated this specific code.

ckarlof:
> Still, since it's not technically the last time the user authed, I'd rather if we named it auth_at. It correctly conveys the time the user had authed with the auth-server when we generated this specific code.

Thoughts, @rfk? I'm good with `auth_at` in the `/token` response.

rfk:
`auth_at` is indeed more to-the-point, thumbs up

rfk:
Amended with @seanmonstar review comments and `last_auth_at` => `auth_at`

lib/db/mysql/patches/patch-000-001.sql
```diff
@@ -0,0 +1,9 @@
+-- Create the 'dbMetadata' table.
+-- Note: This should be the only thing in this initial patch.
+
+CREATE TABLE dbMetadata (
+  name VARCHAR(255) NOT NULL PRIMARY KEY,
+  value VARCHAR(255) NOT NULL
+) ENGINE=InnoDB;
+
+INSERT INTO dbMetadata SET name = 'schema-patch-level', value = '1';
```
seanmonstar: huh, the patch itself is responsible for updating the patch-level? The code looks fine, and won't block this, but I'd personally expect this to be handled by the Patcher itself...

seanmonstar:
r+! i guess i'll wait till we discuss in meeting to merge?

lib/db/mysql/patches/patch-000-001.sql
rfk: The patcher currently sanity-checks that the patch correctly updated the level, but yeah, we should iterate on this

kumar303:
Cool! What's the best server to test this out on? We've been using https://oauth-stable.dev.lcip.org but we can use whatever. I hadn't found time to try it on @rfk's dev server yet.

seanmonstar:
@kumar303 https://oauth-latest.dev.lcip.org tracks master

kumar303:
thanks, I'll try it there

PR #208, opened 2015-01-29
feat(db): add basic migration infrastructure to mysql backend

### LINKED ISSUE #183: DB migration infrastucture

rfk:
This is a work-in-progress branch to import the db-patcher functionality from fxa-auth-db-server.  Not ready for merge.

Fixes #183

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -2,16 +2,19 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const path = require('path');
+
 const buf = require('buf').hex;
 const hex = require('buf').to.hex;
 const mysql = require('mysql');
+const MysqlPatcher = require('mysql-patcher');
 
-const encrypt = require('../encrypt');
-const logger = require('../logging')('db.mysql');
-const P = require('../promise');
-const unique = require('../unique');
+const encrypt = require('../../encrypt');
+const logger = require('../../logging')('db.mysql');
+const P = require('../../promise');
+const unique = require('../../unique');
+const patch = require('./patch');
 
-const SCHEMA = require('fs').readFileSync(__dirname + '/schema.sql').toString();
 
 function MysqlStore(options) {
   if (options.charset && options.charset !== 'UTF8_UNICODE_CI') {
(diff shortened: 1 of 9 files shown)


rfk:
Alrighty @seanmonstar and @chilts, here's an attempt at adding db migrations to this repo, inspired by the way we do it with mysql-patcher in fxa-auth-db-server.  What do you think?

@chilts, you may notice that I'm cheating and using the private `readDbPatchLevel` function from `mysql-patcher` here.  I think it might be worth doing a new release to expose this properly as a public API since "check current patch level" is something we're now doing in two projects.

I also filed https://github.com/chilts/mysql-patcher/issues/10 re: the use of `multipleStatements=true`.

@seanmonstar, please r? modulo the above mysql-patcher tweaks.

rfk:
I'm going to try this out on a dannybox deployment, but I *think* that the patcher should automagically start working with no change, via the following logic:

  * we use `createSchema: true` on the dannyboxes
  * if there's no `dbMetadata` table, then `mysql-patcher` assumes we're at patch level zero and will apply 000-001 and 001-002 in order
  * 000-001 will create the `dbMetadata` table
  * 001-002 will update the patch level but have no other effect, since the tables already exist

For deployment to prod, ops need only run the same two sets of patches.

rfk:
> I'm going to try this out on a dannybox deployment, but I think that the patcher should automagically start working with no change

I completed this today and it went really well - upgraded from current master, to this PR, to the last-auth-at PR, and the patcher automagically did the right thing at each stage.  I love when things work out.

rfk:
note to self: comment on the use of `IF NOT EXISTS` in patch-001-002 - it's good to use here because the tables already exist, but we should discourage any future patches from using it as they should fail noisily if other tables exist.

rfk:
assigning to @seanmonstar for review

rfk:
mysql-patcher API tweaks in https://github.com/chilts/mysql-patcher/pull/11, and this PR updated to use them.  It's sligthly cleaner, but we could iterate further in subsequent versions.

rfk:
dammit, npm-shrinkwrap....

rfk:
The new version of mysql-patcher has been released, and the API changed cleaned this up alot.  @seanmonstar this is now ready for final review.

seanmonstar:
Hmm, it seems that #205 includes this code, right? I left comments about this code in that PR. I can make them again here if it's difficult to correlate them.

rfk:
> Hmm, it seems that #205 includes this code, right?

It does.  Simplest option may be to just close this and complete review of both over in #205

PR #211, opened 2015-02-02
fix(api): reject requests with invalid parameters

### LINKED ISSUE #210: Reject unexpected parameters on most endpoints

seanmonstar:
This keeps the leniency for the POST /v1/token endpoint, but
cactches errors for misuses of other endpoints.

Closes #210 

@washort sanity checking, this doesn't break you, does it? It's only the token endpoint that receives extra parameters?

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -158,7 +158,7 @@ describe('/v1', function() {
     describe('GET', function() {
       it('redirects with all query params', function(done) {
         Server.api
-        .get('/authorization?client_id=123&state=321&scope=1&action=signup')
+        .get('/authorization?client_id=123&state=321&scope=1&action=signup&a=b')
         .then(function(res) {
           assert.equal(res.statusCode, 302);
           var redirect = url.parse(res.headers.location, true);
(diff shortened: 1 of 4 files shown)


test/api.js
```diff
@@ -157,7 +157,7 @@ describe('/v1', function() {
     describe('GET', function() {
       it('redirects with all query params', function(done) {
         Server.api
-        .get('/authorization?client_id=123&state=321&scope=1&action=signup')
+        .get('/authorization?client_id=123&state=321&scope=1&action=signup&a=b')
```
rfk: Maybe worth adding  a comment about what the `a=b` is testing for, and/or a comment as to why we allow `stripUnknown` in the /token case.

rfk:
r+ with a single nit

ckarlof:
So the outcome here is that `GET /authorization` still forwards unanticipated params, but others block?

seanmonstar:
This didn't actually affect the `GET` request, as there was no validation on that route. It's the `POST /token` request that needed that (and so uses `stripUnknown`). All other endpoints will error with a 400.

ckarlof:
Gotcha.

PR #222, opened 2015-02-19
feat(developers): adds support for oauth developers

vladikoff:


vladikoff:
@seanmonstar @rfk can I get some feedback when you get a chance?

lib/db/mysql/schema.sql
```diff
@@ -43,3 +43,14 @@ CREATE TABLE IF NOT EXISTS tokens (
   scope VARCHAR(256) NOT NULL,
   createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
 ) ENGINE=InnoDB CHARACTER SET utf8 COLLATE utf8_unicode_ci;
+
+CREATE TABLE IF NOT EXISTS developers (
+  email VARCHAR(256) NOT NULL PRIMARY KEY,
```
vladikoff: TODO: here I'm planning to add something like `developerId BINARY(32) NOT NULL PRIMARY KEY,` to assign every developer an id. Would that be the best way to set an id in InnoDB?

lib/db/mysql/schema.sql
```diff
@@ -43,3 +43,14 @@ CREATE TABLE IF NOT EXISTS tokens (
...
+  createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+) ENGINE=InnoDB CHARACTER SET utf8 COLLATE utf8_unicode_ci;
+
+CREATE TABLE IF NOT EXISTS clientDevelopers (
+  developerId BINARY(16) NOT NULL,
```
vladikoff: See https://github.com/mozilla/fxa-oauth-server/pull/222/files#r26528695

lib/db/memory.js
```diff
@@ -187,7 +203,54 @@ MemoryStore.prototype = {
...
+    if (! email) {
+      return P.reject(new Error('Email is required'));
+    }
+
+    delete this.developers[email];
+    return P.resolve();
+  },
+
+  registerClientDeveloper:
+  function registerClientDeveloper(developerId, clientId) {
```
vladikoff: I have to bump this to a new line to make sure the line is not too long, any other style ideas or is this fine?

lib/db/mysql/index.js
```diff
@@ -104,6 +104,14 @@ const QUERY_CLIENT_REGISTER =
   'INSERT INTO clients ' +
   '(id, name, imageUri, secret, redirectUri, whitelisted, canGrant) ' +
   'VALUES (?, ?, ?, ?, ?, ?, ?);';
+const QUERY_CLIENT_DEVELOPER_INSERT =
+  'INSERT INTO clientDevelopers ' +
+  '(userId, clientId) ' +
+  'VALUES (?, ?);';
+const QUERY_DEVELOPER_INSERT =
+  'INSERT INTO developers ' +
+  '(email) ' +
```
vladikoff: TODO: add developerId here, part of https://github.com/mozilla/fxa-oauth-server/pull/222/files#r26528695

lib/db/memory.js
```diff
@@ -187,7 +203,54 @@ MemoryStore.prototype = {
...
+
+    return P.resolve();
+
+  },
+  developerFind: function (email) {
+    if (! email) {
+      return P.reject(new Error('Email is required'));
+    }
+
+    return P.resolve(this.developers[email]);
```
vladikoff: Should this Reject or resolve `null`?

lib/db/memory.js
```diff
@@ -187,7 +203,54 @@ MemoryStore.prototype = {
     deleteByUserId(this.tokens, userId);
     deleteByUserId(this.codes, userId);
     return P.resolve();
-  }
+  },
+  developerRegister: function developerRegister(email) {
```
rfk: nit: the verb/noun order on these names is inconsistent with the existing methods in this module, e.g. `removeUser` versus `developerRemove`.

lib/db/mysql/patches/patch-003-004.sql
```diff
@@ -0,0 +1,15 @@
+-- Adds support for Client Developers for OAuth clients
+
+CREATE TABLE IF NOT EXISTS clientDevelopers (
```
rfk: I recommend removing `IF NOT EXISTS` here, so that if someone tries to replay the patch after it has already been applied, we get a noisy failure.  (But keep it in the schema.sql for creating from scratch)

lib/db/mysql/patches/patch-003-004.sql
```diff
@@ -0,0 +1,15 @@
...
+  developerId INT NOT NULL,
```
rfk: Suggest a bytestring id rather than integer id here, for consistency with `clientId` and others.

rfk: heh, I see you're way ahead of my in your coments further down...

lib/db/mysql/schema.sql
rfk: IMHO the simplest approach will be to generate a unique ID in the app and pass it in explicitly when inserting the row.  Using a unique index at the db level would prevent against the very small chance of accidental duplication.

Also, what's the thinking behin 16-versus-32 bytes?

lib/routes/developers/register.js
```diff
@@ -0,0 +1,32 @@
...
+          if (developer) {
+            return reply(developer).code(200);
+          }
+        },
+        reply
+      ).then(function () {
+        return db.developerRegister(email);
+      }).done(function (developer) {
+        return reply(developer).code(200);
+      }, reply);
```
rfk: You may need to add new `errno`'s and document them in the API docs.  For example a "developer already exists" errno.

rfk:
The data model here seems sensible.  How do you plan to expose `registerClientDeveloper` etc functionality to the outside world?  As part of the web API, on the special internal management port?

lib/db/memory.js
rfk: I slightly favour rejecting with an explicit error.

lib/db/memory.js
seanmonstar: Agreed.

lib/db/memory.js
seanmonstar: In all the other db methods, it resolves with `null` (or perhaps `undefined`). I'd say keep it consistent, or change them all (could be a follow up if this is greatly desired).

lib/db/memory.js
seanmonstar: You could shrink the `name` part. I only put that all over the place to improve stack traces.

lib/db/memory.js
```diff
@@ -187,7 +203,54 @@ MemoryStore.prototype = {
...
+    var entry = {
+      developerId: developerId,
+      clientId: clientId,
+      createdAt: new Date()
+    };
+    var uniqueHexId = unbuf(unique.id());
+
+    logger.debug('registerClientDeveloper', entry);
+    this.clientDevelopers[uniqueHexId] = entry;
+    return P.resolve();
```
seanmonstar: Resolve with the `entry` or `id`?

lib/db/mysql/index.js
```diff
@@ -104,6 +104,14 @@ const QUERY_CLIENT_REGISTER =
   'INSERT INTO clients ' +
   '(id, name, imageUri, secret, redirectUri, whitelisted, canGrant) ' +
   'VALUES (?, ?, ?, ?, ?, ?, ?);';
+const QUERY_CLIENT_DEVELOPER_INSERT =
+  'INSERT INTO clientDevelopers ' +
+  '(userId, clientId) ' +
```
seanmonstar: The `userId` here should be `developerId`, as it is in the schema further down.

lib/db/memory.js
rfk: consistency++

lib/db/mysql/schema.sql
seanmonstar: - clientId: 8
- userId: 16
- clientSecret: 32
- code: 32
- token: 32

We could probably get by with `BINARY(8)`, but if we wanted to be extra safe, we could use 16, like userId.

seanmonstar: Separately, we wanted to keep developers by email, instead of `userId` from fxa?

lib/db/mysql/schema.sql
```diff
@@ -43,3 +43,14 @@ CREATE TABLE IF NOT EXISTS tokens (
...
+  clientId BINARY(8) NOT NULL,
```
seanmonstar: I'd add a `FOREIGN KEY` constraint to this column, here and in the patch file.

lib/routes/developers/register.js
```diff
@@ -0,0 +1,32 @@
...
+    scope: [auth.SCOPE_CLIENT_MANAGEMENT]
+  },
+  handler: function developerRegistration(req, reply) {
+    var email = req.auth.credentials.email;
+
+    return db.developerFind(email)
+      .then(
+        function(developer) {
+          if (developer) {
+            return reply(developer).code(200);
```
seanmonstar: Trying to register a developer that already exists is a 200? I'd expect some 400 error, personally...

vladikoff:
@rfk @seanmonstar feedback?

Some parts are missing still (i.e delete should require ownership), going to hook up the oauth console branch tomorrow to see what exactly is left to add.

> How do you plan to expose registerClientDeveloper etc functionality to the outside world? As part of the web API, on the special internal management port?

`activateDeveloper` -> `/developer/activate` is on the internal port. In prod it's also there but I assume we don't need any extra measures.

lib/db/mysql/index.js
```diff
@@ -177,7 +195,56 @@ MysqlStore.prototype = {
       return client;
     });
   },
+  registerClientDeveloper: function regClientDeveloper(developerId, clientId) {
+    if (!developerId || !clientId) {
+      var err = new Error('Owner registration requires user and developer id');
+      return P.reject(err);
+    }
+
+    logger.debug('registerClientOwner: ', developerId, clientId);
```
seanmonstar: This format is illegal for heka. The first argument must be a string of no spaces (probably no colons either). The second argument should be either 1 value, or an object mapping names and values.

lib/db/mysql/patches/patch-003-004.sql
```diff
@@ -0,0 +1,17 @@
+-- Adds support for Client Developers for OAuth clients
+
+CREATE TABLE clientDevelopers (
+  developerId BINARY(16) NOT NULL,
```
seanmonstar: Foreign key this to `developers.developerId` as well.

lib/db/mysql/patches/patch-003-004.sql
```diff
@@ -0,0 +1,17 @@
+-- Adds support for Client Developers for OAuth clients
+
+CREATE TABLE clientDevelopers (
```
seanmonstar: This table needs a primary key.

lib/db/mysql/schema.sql
```diff
@@ -43,3 +43,18 @@ CREATE TABLE IF NOT EXISTS tokens (
   scope VARCHAR(256) NOT NULL,
   createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
 ) ENGINE=InnoDB CHARACTER SET utf8 COLLATE utf8_unicode_ci;
+
+CREATE TABLE IF NOT EXISTS clientDevelopers (
```
seanmonstar: Same here regarding primary key and foreign key.

lib/routes/developers/activate.js
```diff
@@ -0,0 +1,37 @@
...
+          }
+        },
+        reply
+      ).then(function() {
+        return db.activateDeveloper(email);
+      }).then(function() {
+        return db.getDeveloper(email);
+      }).done(function(developer) {
+        developer.developerId = hex(developer.developerId);
+        return reply(developer).code(200);
```
seanmonstar: `.code(200)` is redundant.

test/api.js
```diff
@@ -789,16 +789,55 @@ describe('/v1', function() {
...
+        // this developer has no clients associated, it returns 0
+        // value is the same as the API endpoint and a DB call
+
           return Server.internal.api.get({
             url: '/clients',
             headers: {
               authorization: 'Bearer ' + tok
             }
           }).then(function(res) {
+            console.log(res.result);
```
seanmonstar: Stray console.log

test/db.js
```diff
@@ -142,4 +142,167 @@ describe('db', function() {
...
+              return done();
+            }
+          }, done);
+
+      });
+
+    });
+
+  });
+
```
seanmonstar: Nice! Additional tests are always great!

lib/routes/developers/activate.js
```diff
@@ -0,0 +1,37 @@
...
+            return reply(developer).code(200);
+          }
+        },
+        reply
+      ).then(function() {
+        return db.activateDeveloper(email);
+      }).then(function() {
+        return db.getDeveloper(email);
+      }).done(function(developer) {
+        developer.developerId = hex(developer.developerId);
```
seanmonstar: I _think_ this can poison the JSON DB. It returns the actual object, and modifying a property here will change it in the DB also.

GitCop:

 * Commit: 9328d4c2397f87e7159408d89767e6723b871806

---

vladikoff:
@seanmonstar @rfk r? 

In the next phase of this feature we will be adding endpoints for multiple owners.

The oauth console now lists only the clients you own and should allow you to modify only those: 
devs

test/api.js
```diff
@@ -899,42 +951,115 @@ describe('/v1', function() {
...
       describe('POST /:id', function() {
         var id = unique.id();
-        it('should update the client', function() {
-          var secret = unique.secret();
-          var imageUri = 'https://example.foo.domain/logo.png';
+
+        it('should forbid update to unknown developers', function() {
+          var id = unique.id();
+          var uid = unique(16).toString('hex');
+          var vemail = unique(4).toString('hex') + '@mozilla.com';
```
vladikoff: I think I'm gonna create helper methods for user / tokens and client creation. There is just too much repetition in these tests

lib/db/memory.js
```diff
@@ -187,7 +222,124 @@ MemoryStore.prototype = {
...
+    return P.resolve(entry);
+  },
+  getClientDevelopers: function getClientDevelopers(clientId) {
+    var self = this;
+    var developers = [];
+
+    if (! clientId) {
+      return P.reject(new Error('Client id is required'));
+    }
+
```
seanmonstar: I would add `clientId = unbuf(clientId)` in case a Buffer was passed.

lib/db/memory.js
```diff
@@ -187,7 +222,124 @@ MemoryStore.prototype = {
...
+    return P.resolve();
+  },
+  developerOwnsClient: function devOwnsClient(developerEmail, clientId) {
+    var self = this;
+    var developerId;
+
+    logger.debug('developerOwnsClient');
+    return self.getDeveloper(developerEmail)
+      .then(function (developer) {
+        developerId = developer.developerId;
```
arvindh5686: // developer object returned from getDeveloper might be undefined. Add a check for that.

vladikoff: :+1:

lib/db/memory.js
```diff
@@ -187,7 +222,124 @@ MemoryStore.prototype = {
...
+
+        return self.getClientDevelopers(clientId);
+      })
+      .then(function (developers) {
+        var result = false;
+
+        developers.forEach(function (developer) {
```
seanmonstar: I'd switch to Array.some so that the loop stops once we've found the right developerId.

lib/db/memory.js
```diff
@@ -187,7 +222,124 @@ MemoryStore.prototype = {
...
+          if (developer.developerId === developerId) {
+            result = true;
+          }
+        });
+
+        if (result) {
+          return P.resolve();
```
seanmonstar: The style throughout the rest of the repo is to `resolve` with the state, and only to use `reject` for genuine errors. I'd change to `return P.resolve(result);`

lib/db/memory.js
vladikoff: Done.

vladikoff:
@rfk , @seanmonstar should be target the next train to land this?
I will need to send a mailing list message to let everyone know that their clients in the console will not be visible anymore but will still work.

I'll start working on https://github.com/mozilla/fxa-oauth-server/issues/239 while this is landing.

lib/db/memory.js
```diff
@@ -187,7 +222,128 @@ MemoryStore.prototype = {
...
+      .then(function (developers) {
+        var result;
+
+        function hasDeveloper(developer) {
+          result = developer;
+          return unbuf(developer.developerId) === unbuf(developerId);
+        }
+
+        if (developers.some(hasDeveloper)) {
+          return P.resolve(result);
```
seanmonstar: This returns the developer object, or a boolean? Looks like the developer object to me, though the name of the method sounds like a question asking for a yes or no.

lib/db/mysql/index.js
```diff
@@ -177,7 +200,75 @@ MysqlStore.prototype = {
...
+    return this._write(QUERY_DEVELOPER_DELETE, [
+      email
+    ]);
+  },
+  developerOwnsClient: function devOwnsClient(developerEmail, clientId) {
+    return this._readOne(QUERY_DEVELOPER_OWNS_CLIENT, [
+      developerEmail, buf(clientId)
+    ]).then(function(result) {
+      if (result) {
+        return P.resolve(result);
```
seanmonstar: As above, though it looks like here, the query will return the `rowId`. Whichever result is returned, it should be consistent, or we'll get errors.

lib/routes/developer/activate.js
```diff
@@ -0,0 +1,41 @@
...
+    scope: [auth.SCOPE_CLIENT_MANAGEMENT]
+  },
+  handler: function activateRegistration(req, reply) {
+    var email = req.auth.credentials.email;
+
+    return db.getDeveloper(email)
+      .then(
+        function(developer) {
+          if (developer) {
+            developer.developerId = hex(developer.developerId);
```
seanmonstar: This needs to make a `clone`, because when using the memory database, this will mutate the value in the db as well.

lib/routes/developer/activate.js
```diff
@@ -0,0 +1,41 @@
...
+            return reply(developer).code(200);
```
seanmonstar: Also, I'm confused by this return here. It will return a `reply` object, so the promise chain will continue, right? Meaning the db will get queried and mutated after the response of 200 was sent.

test/api.js
```diff
@@ -130,6 +130,33 @@ function assertRequestParam(result, param) {
   assert.equal(result.validation.keys[0], param);
 }
 
+// helper function to create a new user, email and token for some client
+/**
+ *
+ * @param {String} cId - hex client id
+ */
+function getUniqueUserAndToken(cId) {
```
seanmonstar: woo!

lib/routes/developer/activate.js
```diff
@@ -0,0 +1,49 @@
...
+            return reply(developerResponse(developer)).code(200);
+          } else {
+            return db.activateDeveloper(email)
+              .then(function() {
+                return db.getDeveloper(email);
+              })
+              .done(function(developer) {
+                if (! developer) {
+                  var msg = 'Developer activation failed';
+                  return reply(AppError.unauthorized(msg));
```
vladikoff: @seanmonstar Thanks for the reviews. What type of error do you prefer here? This will happen if activation fails in the earlier step.

vladikoff: @rfk suggested to use a 500 here. will update

vladikoff:
ping @seanmonstar

lib/routes/client/register.js
```diff
@@ -50,16 +50,34 @@ module.exports = {
...
-    }, reply);
+    var developerEmail = req.auth.credentials.email;
+    var developerId = null;
+
+    return db.getDeveloper(developerEmail)
+      .then(function (developer) {
+
+        // must be a developer to register clients
+        if (! developer) {
+          return reply(AppError.unauthorized('Illegal Developer'));
```
seanmonstar: I'd change this to `throw AppError.unauthorized`. This will make it skip all the other `then` paths, and just use the last error path, which at L81 will `reply` with the error. Otherwise, since this returns the promise with a non error, it will execute the rest of the `then`s.

lib/routes/developer/activate.js
```diff
@@ -0,0 +1,52 @@
...
+            return reply(developerResponse(developer)).code(200);
+          } else {
+            return db.activateDeveloper(email)
+              .then(function() {
+                return db.getDeveloper(email);
+              })
+              .done(function(developer) {
+                // check to make sure activation succeeded
+                if (! developer) {
+                  return reply(new AppError({
```
vladikoff: @seanmonstar same as https://github.com/mozilla/fxa-oauth-server/pull/222/files#r29173658 ?

seanmonstar: I'd personally restructure this one to this:

```js
return db.getDeveloper(email)
  .then(function(developer) {
    if (developer) {
      return developer;
    } else {
      // better to just return the developer here as well, instead of using an additional query
      return db.activeDeveloper(email);
    }
  })
  .then(developerResponse)
  .done(reply, reply);
```

seanmonstar: And if there was a failure activating the developer, and an error was thrown, that would bubble the error up to the error reply case.

lib/db/mysql/index.js
```diff
@@ -177,7 +200,77 @@ MysqlStore.prototype = {
...
+      return P.reject(new Error('Email is required'));
+    }
 
+    var developerId = unique.developerId();
+    logger.debug('activateDeveloper', { developerId: developerId });
+    return this._write(QUERY_DEVELOPER_INSERT, [
+      developerId, email
+    ]).then(function () {
+      return this.getDeveloper(email);
+    }.bind(this));
```
vladikoff: @seanmonstar added a query here.

vladikoff:
updated.

seanmonstar:
Let's try it!

PR #223, opened 2015-02-19
fix(clients): update email validation

vladikoff:


vladikoff:
Thanks @seanmonstar !

PR #225, opened 2015-02-19
chore(clients): add credentials for FF/FFOS/Fennec/FxA clients in dev

zaach:
Adds clients from: https://bugzilla.mozilla.org/show_bug.cgi?id=1064505

This will be useful for devs working on these clients locally. I'm not sure if these are available on latest but we might want to create them there, too.

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -41,6 +41,42 @@
       "hashedSecret": "3c32099123471ffb80a7553558e6a6e8589da2235a4e081fe9d76648aa25d050",
       "whitelisted": true,
       "canGrant": false
+    },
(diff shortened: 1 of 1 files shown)


rfk:
They should exist on all fxa-dev environments: https://github.com/mozilla/fxa-dev/blob/master/roles/oauth/templates/config.json.j2#L189

zaach:
@rfk Ah cool, they're already in fxa-dev :+1:

rfk:
This seems fine to me, not sure what travis is up to but it can't possibly be related to your changes here, right..?

PR #227, opened 2015-03-03
feat(clients): move client management api to a separate port

seanmonstar:
This will allow the management APIs to be turned on in "production", but using a port that is not publicly accessible.

/cc @ckolos @ckarlof

### DIFF
--- a/bin/internal.js
+++ b/bin/internal.js
@@ -4,16 +4,14 @@
 
 const config = require('../lib/config').root();
 const db = require('../lib/db');
-const logger = require('../lib/logging')('bin.server');
-const server = require('../lib/server').create();
-const events = require('../lib/events');
+const logger = require('../lib/logging')('bin.internal');
+const server = require('../lib/server/internal').create();
 
 logger.debug('config', config);
 db.ping().done(function() {
   server.start(function() {
     logger.info('listening', server.info.uri);
   });
-  events.start();
 }, function(err) {
   logger.critical('db.ping', err);
 });
--- a/lib/routing.js
+++ b/lib/routing.js
@@ -9,7 +9,7 @@ function v(url) {
   return '/v' + version + url;
 }
 
-module.exports = [
+exports.routes = [
   {
     method: 'GET',
     path: '/',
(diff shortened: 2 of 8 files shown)


bin/internal.js
```diff
@@ -0,0 +1,17 @@
...
+
+const config = require('../lib/config').root();
+const db = require('../lib/db');
+const logger = require('../lib/logging')('bin.internal');
+const server = require('../lib/server/clients').create();
+
+logger.debug('config', config);
+db.ping().done(function() {
+  server.start(function() {
+    logger.info('listening', server.info.uri);
```
rfk: We could include the name of the server in this log line, as a debugging aid to distinguish bin/server.js and bin/internal.js

rfk: Oh, actually I guess this will come out for free because we pass the name to the logger, never mind :-)

lib/routing.js
```diff
@@ -59,27 +59,25 @@ module.exports = [
...
-      config: require('./routes/client/update')
-    },
-    {
-      method: 'DELETE',
-      path: v('/client/{client_id}'),
-      config: require('./routes/client/delete')
-    }
-  ]);
-}
+exports.clients = [
```
rfk: naming nit: this is a list of routes, so it would be clearer as something like "clientRoutes" or "internalRoutes".

lib/server/clients.js
```diff
@@ -0,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: Should we call this file "internal.js" for consistency with "bin/internal.js"?

test/api.js
```diff
@@ -746,244 +746,257 @@ describe('/v1', function() {
...
-          assert.equal(res.statusCode, 401);
+    describe('client management api', function() {
+      it('should not be available on main server', function(){
+        return P.all([
+          Server.api.get('/clients'),
+          Server.api.post('/client'),
+          Server.api.post('/client/' + clientId),
+          Server.api.delete('/client/' + clientId)
+        ]).map(function(res) {
+          assert.equal(res.statusCode, 404);
```
rfk: Wow, the diff really made a mess of this file!  Is this mostly just shuffling the existing tests so that they point to the appropriate server, or should I be reviewing anything here in more detail?

lib/server/clients.js
```diff
@@ -0,0 +1,93 @@
...
+      security: {
+        hsts: {
+          maxAge: 15552000,
+          includeSubdomains: true
+        },
+        xframe: false,
+        xss: false,
+        noOpen: false,
+        noSniff: false
+      }
```
rfk: It may be worth factoring this server config out into a shared constant.

rfk:
Good stuff @seanmonstar, looking forward to putting this into production.  Some style suggestions but otherwise r+.

test/api.js
seanmonstar: Yes. There was also one level of indentation added, so appending `?w=1` to the PR url will make it easier to see.

rfk: nice! TIL...

lib/server/clients.js
```diff
@@ -0,0 +1,93 @@
...
+
+  var routes = require('../routing').clients;
+  if (isProd) {
+    logger.info('prod', 'Disabling response schema validation');
+    routes.forEach(function(route) {
+      delete route.config.response;
+    });
+  }
+
+  // require json by default
```
jrgm: Comment is out-of-sync with `allow: ['application/json', 'application/x-www-form-urlencoded']`

test/api.js
rfk: So yeah, just to confirm that this all looks find with the `?w=1` on the URL.

lib/routing.js
seanmonstar: I see the name here as `routing.clients`, since the name of the module.

lib/server/clients.js
seanmonstar: Probably.

seanmonstar:
Fixed style issues, merged.

PR #237, opened 2015-03-27
chore(dependencies): update most dependencies

seanmonstar:
Hapi will be separate, since it is quite a breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,37 +23,37 @@
     "node": ">=0.10.0"
   },
   "dependencies": {
-    "URIjs": "1.14.1",
-    "bluebird": "2.3.10",
+    "URIjs": "^1.14.2",
+    "bluebird": "^2.9.14",
     "buf": "0.1.0",
     "convict": "0.6",
     "fxa-notifier-aws": "1.0.0",
     "hapi": "7.5.3",
-    "joi": "4.7.0",
-    "mozlog": "1.0.0",
-    "mysql": "2.5.2",
-    "mysql-patcher": "0.6.0",
-    "request": "2.47.0"
+    "joi": "^4.7.0",
+    "mozlog": "^2.0.0",
+    "mysql": "^2.5.5",
+    "mysql-patcher": "^0.7.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -23,37 +23,37 @@
     "node": ">=0.10.0"
   },
   "dependencies": {
-    "URIjs": "1.14.1",
-    "bluebird": "2.3.10",
+    "URIjs": "^1.14.2",
```
vladikoff: All these `^` should probably be removed. All other repos have this locked to avoid update surprises.

seanmonstar: The shrinkwrap should take care of it, no?

vladikoff: I think it won't take care of it if someone adds a new (dev)dependency and rewraps it.

seanmonstar: My understanding of shrinkwrap is that it wraps the versions that are
**currently** installed in the node_modules directory. So running it again
at a later time won't adjust the shrinkwrap, other then if you've
intentionally `npm install foo@newer`.

vladikoff: Yeah but if it is a new machine / new developer who just installed a new copy of the server?

seanmonstar: Then `npm install` should obey the shrinkwrap and ignore the package.json completely.

vladikoff: Okay, @pdehaan might also know why we chose to keep the deps and in some cases devDeps in other servers locked.

pdehaan: No, I'm the stupid one. You're thinking of @jrgm (the smart one).
I thought we did `npm shrinkwrap --dev` in all the repos. We should probably standardize on one method and use it consistently across projects.

jrgm:
My view is that shrinkwrap should be done with `--dev`. My reasoning is that if you want to check out an old version of your project, if you don't lock `--dev`, you may find it doesn't build anymore.

jrgm:
As for the `^`, I think we're good with the shrinkwrap in place.

jrgm:
r+ from me, fwiw

vladikoff:
:+1:

PR #242, opened 2015-04-20
Pool sizes

seanmonstar:
/cc @ckolos

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -130,6 +130,10 @@ const conf = convict({
     port: {
       default: '3306',
       env: 'MYSQL_PORT'
+    },
+    connectionLimit: {
+      doc: 'The maximum number of connections that the pool can use at once.',
+      default: 10
     }
   },
   publicUrl: {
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -130,6 +134,10 @@ const conf = convict({
     port: {
       default: '3306',
       env: 'MYSQL_PORT'
+    },
+    connectionLimit: {
```
vladikoff: just browsing through the PRs, I could not find where this `connectionLimit` is used

seanmonstar: This whole `mysql` config object is passed to `mysql.createPool`, so it will be one of the options to the constructor.

vladikoff: Ah okay, both this and `maxSockets` had the same doc `The maximum number of connections that the pool can use at once.` so it was a bit confusing

rfk:
LGTM, r+

ckolos:
Not sure I understand why the verification configs are removed in the browserid.js file. Were they blocking something for the db connections (seems quite unlikely, but wanted to ask anyway....)

seanmonstar:
@ckolos they weren't, the code was just moved around. I created a `request` instance with some values set as default: the `pool` and the `url`, since they never change. The only thing that changes per request is the JSON payload.

PR #246, opened 2015-04-27
feat(auth): redirect to content-server oauth root by default

### LINKED ISSUE #245: When no action is specified, redirect to a generic authentication route
The generic content-server route will direct the user to the appropriate page (sign in or sign up) depending on their sessions state.

Ref: https://github.com/mozilla/fxa-content-server/issues/2344

zaach:
Fixes #245.

Depends on implementation of https://github.com/mozilla/fxa-content-server/issues/2344

### DIFF
--- a/lib/routes/redirect.js
+++ b/lib/routes/redirect.js
@@ -11,9 +11,11 @@ function actionToPathname(action) {
     return 'signup';
   } else if (action === 'force_auth') {
     return 'force_auth';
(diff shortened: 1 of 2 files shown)


seanmonstar:
"do not merge" because it will currently break content-server?

otherwise, r+

zaach:
@seanmonstar Yeep.

PR #248, opened 2015-04-28
feat(untrusted-clients): restrict scopes that untrusted clients can request

### LINKED ISSUE #243: Restrict scopes for untrusted reliers
We may want to restrict scopes for certain reliers, e.g. a third party relier that can only request the `profile` scope.

Related #143.

zaach:
@rfk r?

Fixes #243.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -212,4 +212,15 @@ AppError.invalidContentType = function invalidContentType() {
   });
 };
 
+AppError.invalidScopes = function invalidScopes(scopes) {
+  return new AppError({
+    code: 400,
+    error: 'Invalid scopes',
+    errno: 114,
+    message: 'Invalid scopes'
+  }, {
+    invalidScopes: scopes
+  });
+};
+
 module.exports = AppError;
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -4,7 +4,6 @@
 
 const buf = require('buf').hex;
 const hex = require('buf').to.hex;
-const Hapi = require('hapi');
 const Joi = require('joi');
 const URI = require('URIjs');
 
@@ -19,6 +18,12 @@ const verify = require('../browserid');
 const CODE = 'code';
 const TOKEN = 'token';
 
+const UNTRUSTED_CLIENT_ALLOWED_SCOPES = [
+  'profile:uid',
+  'profile:email',
+  'profile:display_name'
+];
+
 /*jshint camelcase: false*/
 
 function set(arr) {
@@ -29,15 +34,30 @@ function set(arr) {
   return Object.keys(obj);
 }
 
-
+function scopeStringToSet(scopes) {
+  return typeof scopes === 'string' ?
+    set(scopes.split(/\s+/g)) :
+    scopes;
+}
 
 function isLocalHost(url) {
   var host = new URI(url).hostname();
   return host === 'localhost' || host === '127.0.0.1';
 }
 
-function generateCode(claims, client, scope, req) {
+function detectInvalidScopes(requestedScopes, validScopes) {
+  var invalidScopes = [];
 
+  requestedScopes.forEach(function(scope) {
+    if (validScopes.indexOf(scope) === -1) {
+      invalidScopes.push(scope);
+    }
+  });
+
+  return invalidScopes;
+}
+
+function generateCode(claims, client, scope, req) {
   return db.generateCode(
     client.id,
     buf(claims.uid),
@@ -127,6 +147,7 @@ module.exports = {
     var start = Date.now();
     var wantsGrant = req.payload.response_type === TOKEN;
     var exitEarly = false;
+    var scope = req.payload.scope ? scopeStringToSet(req.payload.scope) : [];
     P.all([
       verify(req.payload.assertion).then(function(claims) {
         logger.info('time.browserid_verify', { ms: Date.now() - start });
@@ -146,9 +167,12 @@ module.exports = {
           logger.debug('notFound', { id: req.payload.client_id });
           throw AppError.unknownClient(req.payload.client_id);
         } else if (!client.whitelisted) {
-          logger.error('notWhitelisted', { id: req.payload.client_id });
-          // TODO: implement external clients so we can remove this
-          throw Hapi.error.notImplemented();
+          var invalidScopes = detectInvalidScopes(scope,
+                                UNTRUSTED_CLIENT_ALLOWED_SCOPES);
+
+          if (invalidScopes.length) {
+            throw AppError.invalidScopes(invalidScopes);
+          }
         }
 
         var uri = req.payload.redirect_uri || client.redirectUri;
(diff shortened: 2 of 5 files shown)


lib/routes/authorization.js
```diff
@@ -19,6 +18,13 @@ const verify = require('../browserid');
 const CODE = 'code';
 const TOKEN = 'token';
 
+const UNTRUSTED_CLIENT_SCOPES = [
```
rfk: nit: the word-order in this name is confusing, it suggests the scopes themselves are somehow untrusted.  Something like `UNTRUSTED_CLIENT_ALLOWED_SCOPES` may be a little clearer.

lib/routes/authorization.js
```diff
@@ -19,6 +18,13 @@ const verify = require('../browserid');
...
+  'profile:uid',
+  'profile:email',
+  'profile:avatar',
```
rfk: In discussion with @ryanfeeley he felt pretty strongly that we should not share the avatar by default.

lib/routes/authorization.js
```diff
@@ -29,15 +35,30 @@ function set(arr) {
   return Object.keys(obj);
 }
 
-
+function scopeStringToSet(scopes) {
+  return typeof scopes === 'string' ?
+    set(scopes.split(' ')) :
```
rfk: It may be worth doing `.split(/\s+/)` here to avoid weirdness if they accidentally pass multiple spaces in the string.

lib/routes/authorization.js
```diff
@@ -29,15 +35,30 @@ function set(arr) {
...
+    scopes;
+}
 
 function isLocalHost(url) {
   var host = new URI(url).hostname();
   return host === 'localhost' || host === '127.0.0.1';
 }
 
-function generateCode(claims, client, scope, req) {
+function scopeCheck(requestedScopes, validScopes) {
```
rfk: nit: I suggest naming this something a little more specific, since to me "scope check" means "check whether a particular scope is included in this list of scopes" rather than "check these scopes for validity".

test/api.js
```diff
@@ -244,6 +250,38 @@ describe('/v1', function() {
...
+          payload: authParams({
+            client_id: client.id,
+            scope: 'profile:email profile:uid'
+          })
+        }).then(function(res) {
+          assert.equal(res.statusCode, 200);
+        });
+      });
+    });
+
```
rfk: It's probably worth adding a test for a multi-scope request like "profile profile:write" or similar, just to be doubly sure.

rfk:
Thanks @zaach, this is looking really solid.

I'm starting to question my earlier point about error out if there's invalid scope, versus just trimming it down to the scopes we allow.  We'll need to make sure that we don't suddenly break pocket's integration when this thing goes live.

If we're happy with doing it this way, let's reach out to them and get them to starting requesting the more restriced scopes now, before merging this to any dev environments.  IIUC they're currently just requesting `profile` scope, but it should all work if they switch to the more restricted variants.

lib/routes/authorization.js
zaach: How about `display_name`?

rfk: I think `display_name` is fine; @ryanfeeley thoughts?

lib/error.js
```diff
@@ -212,4 +212,15 @@ AppError.invalidContentType = function invalidContentType() {
   });
 };
 
+AppError.invalidScopes = function invalidScopes(scopes) {
+  return new AppError({
+    code: 400,
+    error: 'Invalid scopes',
+    errno: 114,
+    message: 'Scopes for untrusted clients are restricted'
```
seanmonstar: I'd like to use this error as well for refresh tokens, when you originally have a scope of `profile:email` and then try to narrow to scopes you don't have, like `profile:avatar`. Could the message be more general?

lib/routes/authorization.js
seanmonstar: "By default"? At least in the context of this PR, until a permissions page is built into the UI, any scope in this list is plain not allowed, default or otherwise.

Does that mean we don't Pocket an avatar until we've added a permissions page?

rfk: Indeed, "by default" has no meaning here.  We had some chat about sending a custom avatar just for pocket, but that's not in scope.

So yes, pocket will not get an avatar until we add a permissions page or decide to adjust this allowlist.

lib/error.js
zaach: :+1:

zaach:
Updated :+1:

rfk:
I just realized that we may need to block giving pocket prod oauth credentials until this lands in prod.  That could seriously affect the timing of https://bugzilla.mozilla.org/show_bug.cgi?id=1157578.

@jrgm, how much of a pain would it be to either sneak this into train-36, or to do a train-36.1 with it included?

lib/routes/authorization.js
```diff
@@ -146,9 +167,12 @@ module.exports = {
...
           throw AppError.unknownClient(req.payload.client_id);
         } else if (!client.whitelisted) {
-          logger.error('notWhitelisted', { id: req.payload.client_id });
-          // TODO: implement external clients so we can remove this
-          throw Hapi.error.notImplemented();
+          var invalidScopes = detectInvalidScopes(scope,
+                                UNTRUSTED_CLIENT_ALLOWED_SCOPES);
+
+          if (invalidScopes.length) {
+            return P.reject(AppError.invalidScopes(invalidScopes));
```
rfk: nit: the nearby code just throws rather than returning a rejected promise, we should be consistent

lib/error.js
```diff
@@ -212,4 +212,15 @@ AppError.invalidContentType = function invalidContentType() {
...
+  return new AppError({
+    code: 400,
+    error: 'Invalid scopes',
+    errno: 114,
+    message: 'Invalid scopes'
+  }, {
+    invalidScopes: scopes
+  });
+};
+
```
rfk: Please add this new error code to ./docs/API.md (or file a follow-up bug to do it once the pace slows down a little)

rfk:
I'm r+ on this, nice work @zaach.  I'll reach out to pocket and ask them to start requesting only the allowed scopes, and we can merge this once we settle on a timeline for getting it into prod.

jrgm:
> @jrgm, how much of a pain would it be to either sneak this into train-36, or to do a train-36.1 with it included?

Yeah, might as well get this in now. This should be tag 0.36.1, to keep things clear.

rfk:
Re: doing a point-release, we can probably avoid it for now.  I've temporarily provisioned pocket as a whitelisted client in prod/stage for testing purposes, on the theory that (a) they're contractually obliged to do the right thing and (b) they can't do much damage anyway, apart from sneakily reading extra profile information.  So we can take a bit more time to land this and related API changes before asking them to switch over.

rfk:
That said, I'm pretty sure we're safe to land this on master now.  I'm r+, @zaach to coordinate the merge.

Issue #231, opened 2015-03-09
Remove obsolete generate-client.js script


rfk:
We're now doing this via a command-line client to a web api, so the script is no longer necessary.

PR #251, opened 2015-04-29
DB changes in preparation for third-party reliers

### LINKED ISSUE #188: Change "whitelisted" relier configuration flag to "trusted", so that it's more meaningful and less biased
These things matter:

http://www.cs.virginia.edu/~evans/cs655/readings/purity.html

### LINKED ISSUE #249: Add extra client meta-data for third-party reliers
Tracking bug to add additional client meta-data in service of https://github.com/mozilla/fxa-content-server/issues/2347 and https://github.com/mozilla/fxa-content-server/issues/2346.  We'll need at least ToS/PP links for each relier.  @zaach can you think of any other fields we will need?

rfk:
Here are a couple of commits that make DB changes in preparation for third-party untrusted reliers.  I'm putting them in a single PR in order to produce a single DB migration, but it's really two separate ideas.

The first change is the search-replace of "whitelisted" with "trusted" in the client metadata, as discussed in #188.  I've tried to do it in a deployment-friendly and backwards-compatible way, e.g. it still accepts "whitelisted" as input when registering a client, but canonicalized to "trusted" internally.  In theory, existing tooling around this will continue to work.  We'll need matching bugs to update oauth-console and oauth-client, and a follow-on migration to drop the "whitelisted" column.

In addition, we now expose "trusted" in the client metadata fetched by the content-server, allowing it to e.g. trigger the secondary permissions prompt.

The second change is to add "terms_uri" and "privacy_uri" fields on the client metadata per #249, in a way pretty much entirely analogous to "image_uri".

We should hold off merging this until we decide if we need any extra client metadata for #249, but this is otherwise ready for review, assuming I can convince travis to go green.  @seanmonstar r?

Fixes #188 
Fixes #249


scripts/generate-client.js
```diff
@@ -91,7 +91,7 @@ r({
 
   var secret = unique.secret();
   client.hashedSecret = hash(secret).toString('hex');
-  client.whitelisted = true;
+  client.trusted = true;
```
seanmonstar: Ah yes, this reminds me, we should probably kill this script.

seanmonstar:
Excellent PR! I expect this conflicts with #243, so I'm not sure which we want to merge first...

rfk:
Let's land #243 first, then I'll rebase on top of it.

scripts/generate-client.js
rfk: indeed, #231 is on my pile but I havent' gotten round to it

rfk:
rebased

PR #255, opened 2015-05-13
WIP fix(docs): document keys and verification_redirect options

vladikoff:
@shane-tomlinson @rfk r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -292,6 +292,11 @@ content-server page.
     signed in email address will be used as the default.
   - If `action` is `force_auth`, the user is unable to modify the email
     address and is unable to sign up if the address is not registered.
+- `keys`: Optional. Boolean setting, set this if the relier wants access to the account encryption keys.
+- `verification_redirect`: Optional. This option adds a "Proceed" button into the "Account Ready" view. See options for details.
+  - Default. If `verification_redirect` is `no` the account ready view will not show a "Proceed" button that will return to the relier.
+  - If `verification_redirect` is `samebrowser` the account ready view will show a "Proceed" to the relier button only if the flow is in the same browser.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -292,6 +292,11 @@ content-server page.
     signed in email address will be used as the default.
   - If `action` is `force_auth`, the user is unable to modify the email
     address and is unable to sign up if the address is not registered.
+- `keys`: Optional. Boolean setting, set this if relier will need derive relier keys during the signup / signin flow.
```
shane-tomlinson: `derive`=>`derived`. `during the signup / signin flow` may be implicit and not needed?

docs/api.md
```diff
@@ -292,6 +292,11 @@ content-server page.
...
+- `verification_redirect`: Optional. This option adds a "Proceed" button into the "Account Ready" view. See options for details.
```
shane-tomlinson: This is difficult to document too. A "Proceed" button is only displayed if the user verifies in a 2nd browser. If the user verifies in the same browser, they are automatically redirected w/o user interaction.

vladikoff:
Missing feedback, adding
image

docs/api.md
```diff
@@ -292,6 +292,11 @@ content-server page.
     signed in email address will be used as the default.
   - If `action` is `force_auth`, the user is unable to modify the email
     address and is unable to sign up if the address is not registered.
+- `keys`: Optional. Boolean setting, set this if the relier wants access to the account encryption keys.
```
rfk: IMO "wants access to the account encryption keys" suggests they can access the raw keys, which is not the case.  Something more generic like "wants to receive encryption keys" might be better.

vladikoff:
Setting this as WIP for now, until we get the final version of verification_redirect

vladikoff:
Will update!

vladikoff:
Added via https://github.com/mozilla/fxa-oauth-server/commit/ef8c47a5a7f424e60aa218516083a8fb178a8645

PR #256, opened 2015-05-14
fix(api): ensure /destroy endpoint returns an empty object in response body

### LINKED ISSUE #236: /v1/destroy returns mysql result object in JSON response
I noticed this while mucking around on the command-line:
```
$ curl -X POST -H 'Content-Type: application/json' https://oauth.accounts.firefox.com/v1/destroy -d '{"token": "01f09fade2c8504f25d4e50631fce1d6d3fd0a8e93071d039b704d41973b0cad"}'
{"fieldCount":0,"affectedRows":1,"insertId":0,"serverStatus":2,"warningCount":0,"message":"","protocol41":true,"changedRows":0}
```

rfk:
Fixes #236.  (And with apologies to @chilts for stealing his bug; I've been starting at oauth server response bodies for that last hour and this was really starting to grate on me...)

### DIFF
--- a/lib/routes/destroy.js
+++ b/lib/routes/destroy.js
@@ -28,7 +28,8 @@ module.exports = {
         throw AppError.invalidToken();
       }
       return db.removeToken(token);
-    }).done(reply, reply);
-
+    }).done(function() {
+      reply({});
+    }, reply);
   }
 };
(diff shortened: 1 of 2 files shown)


chilts:
@rfk no worries. I've been looking at other stuff. :) This was on my next but one list!

chilts:
Actually, funnily enough I looked at this last week and ended up with the same solution (pretty easy I know) but with `reply.bind({})` instead. I never checked it in or pushed. :( Apologies.

PR #259, opened 2015-05-23
chore(license): Update license to be SPDX compliant

pdehaan:
specifying the type and URL is deprecated:

https://docs.npmjs.com/files/package.json#license


vladikoff:
Thanks Peter!

PR #260, opened 2015-05-26
feat(refresh_tokens): add refresh_tokens to /token endpoint

### LINKED ISSUE #209: Allow clients to request limited lifetime access_tokens
I propose we add an parameter in https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1authorization to do this.

The main use case I have in mind is for Reading List in Desktop to address some kind of revocation. IF the user loses her device, the only mechanism she has to revoke things is password reset. Reading List will be using the implicit grant flow to get a token for the logged in user on Desktop and it would be nice if it could request limited lifetime token. Then when the user resets her password, her FxA session token will expire and her browser's token will no longer be able to mint BiD assertions and get new OAuth tokens. 

Otherwise, we'll have to just request a new token for each request (or at startup), and fill our DB with orphaned tokens. :)

@seanmonstar and @rfk, nothing urgent, but it would be nice to address this in the next 4 weeks or so.

seanmonstar:
All good to go-ish.

Any test cases I missed with scopes?

Fixes #209


GitCop:

 * Commit: e68f74dad1a81b77c5d3e877acf5ebc0fad01923

---

lib/db/mysql/index.js
```diff
@@ -148,13 +151,20 @@ const QUERY_CODE_INSERT =
   'INSERT INTO codes (clientId, userId, email, scope, authAt, code) ' +
   'VALUES (?, ?, ?, ?, ?, ?)';
 const QUERY_TOKEN_INSERT =
-  'INSERT INTO tokens (clientId, userId, email, scope, type, token) VALUES ' +
-  '(?, ?, ?, ?, ?, ?)';
+  'INSERT INTO tokens (clientId, userId, email, scope, type, expiresAt,token)' +
```
pdehaan: Add space before `token`?

lib/db/mysql/index.js
```diff
@@ -148,13 +151,20 @@ const QUERY_CODE_INSERT =
...
+  ' VALUES (?, ?, ?, ?, ?, ?)';
```
pdehaan: Is this missing a `"?"`? It seems like we added `token` above and have 7 fields, but only 6 `?`s.

GitCop:

 * Commit: e68f74dad1a81b77c5d3e877acf5ebc0fad01923

 * Commit: 30c656c0bf6002ef6762a227526f1bbceea03cf6

---

docs/api.md
```diff
@@ -362,7 +362,26 @@ particular user.
 
 - `client_id`: The id returned from client registration.
 - `client_secret`: The secret returned from client registration.
-- `code`: A string that was received from the [authorization][] endpoint.
+- `ttl`: (optional) Seconds that this access_token should be valid.
```
vladikoff: nit: `Seconds that this access_token should be valid.` -> 

`The amount of time the access_token is valid for in seconds` or 
`The auth_token expiration time in seconds`

lib/config.js
```diff
@@ -80,10 +80,15 @@ const conf = convict({
     }
   },
   expiration: {
-    code: {
+    accessTokenS: {
```
vladikoff: nit: `expiration.accessTokenInSeconds`? is this format used anywhere else? (adding S and Ms)?

seanmonstar: I don't know if we use this convention anywhere else. I think we try to use milliseconds whenever we can, but the spec claims that the `expires_at` value should be in seconds...

lib/db/memory.js
```diff
@@ -215,19 +233,41 @@ MemoryStore.prototype = {
...
   },
+  generateRefreshToken: function generateRefreshToken(vals) {
+    var token = {};
+    token.clientId = vals.clientId;
+    token.userId = vals.userId;
+    token.email = vals.email;
+    token.scope = vals.scope;
+    token.createdAt = new Date();
+    var _token = unique.token();
+    var ret = clone(token);
```
vladikoff: nit: group vars with the top `var token = {}`. rename `var token = {};` to something like `tokenData = {}` to avoid `_token` vs `token`

lib/error.js
```diff
@@ -217,7 +217,7 @@ AppError.invalidScopes = function invalidScopes(scopes) {
     code: 400,
     error: 'Invalid scopes',
     errno: 114,
-    message: 'Invalid scopes'
+    message: 'Scopes requested not allowed'
```
vladikoff: Nit: `Scopes requested not allowed` -> i.e `Requested scopes are not allowed`

lib/db/mysql/index.js
```diff
@@ -407,6 +457,7 @@ MysqlStore.prototype = {
     // TODO this should be a transaction or stored procedure
```
vladikoff: Should we file a "good first bug" for this?

lib/routes/token.js
```diff
@@ -11,21 +12,108 @@ const config = require('../config');
...
+    }
+    if (access.expiresAt) {
+      json.expires_at = access.expiresAt;
+    }
+    if (refresh) {
+      json.refresh_token = refresh.token.toString('hex');
+    }
+    return json;
+  });
+  //return [code.authAt, db.removeCode(code.code), db.generateToken(code)];
```
vladikoff: old code from `generateToken` left here.

docs/api.md
```diff
@@ -362,7 +362,26 @@ particular user.
...
+  property in the response reminding of this eventual change.
+- `grant_type`: Either the string `authorization_code` or `refresh_token`.
+  - If `authorization_code`:
+    - `code`: A string that was received from the [authorization][] endpoint.
+  - If `refresh_token`:
+    - `refresh_token`: A string that received from the [token][]
+      endpoint specifically as a refresh token.
+    - `scope`: (optional) A subset of scopes provided to this
+      refresh_token originally, to receive an access_token with less
+      permissions.
```
rfk: > refresh tokens will be mandatory

IIUC there are cases in the OAuth spec where we're *forbidden* from delivering refresh tokens, e.g. when using implicit grant clients.  Perhaps we can weaken this to something like "refresh tokens will be teh default behaviour".

seanmonstar: Woops, this was the first thing I wrote in this PR. It didn't reflect that changes that we'd always expire the token now. Updated the text.

lib/db/mysql/index.js
seanmonstar: I'm not sure if it'd be an "easy" bug...

seanmonstar:
woo, build finally passes

docs/api.md
```diff
@@ -362,7 +362,18 @@ particular user.
...
+  
+  The default and maximum value is 2 weeks.
+- `grant_type`: Either the string `authorization_code` or `refresh_token`.
+  - If `authorization_code`:
+    - `code`: A string that was received from the [authorization][] endpoint.
+  - If `refresh_token`:
+    - `refresh_token`: A string that received from the [token][]
+      endpoint specifically as a refresh token.
+    - `scope`: (optional) A subset of scopes provided to this
+      refresh_token originally, to receive an access_token with less
```
rfk: I was going to go all grammar pedant on this and suggest s/less/fewer/, but Wikipedia tells me [1] that it's actually fine to use "less" with count nouns and I should learn to relax and enjoy life without worrying about made-up grammatical rules dating from 1770...

[1] https://en.wikipedia.org/wiki/Fewer_vs._less

docs/api.md
```diff
@@ -384,6 +397,10 @@ A valid request will return a JSON response with these properties:
 
 - `access_token`: A string that can be used for authorized requests to service providers.
 - `scope`: A string of space-separated permissions that this token has. May differ from requested scopes, since user can deny permissions.
+- `refresh_token`: (Optional) A refresh token to fetch a new access
+  token when this one expires. Only will be present if
+  `grant_type=authorization_code`.
+- `expires_at`: **Millisecond** timestamp when this access token will no longer be valid.
```
rfk: Mixing seconds and milliseconds in this response seems like a recipe for trouble.

More broadly, I think it makes sense to copy the behaviour of existing oauth and/or openid-connect servers where there's precedent.  In this case both specs seem to suggest using `expires_in` with a relative timestamp in seconds, e.g:  http://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse

lib/config.js
rfk: I like millisecond timestamps too, but I think it will be against the grain when interfacing with existing OAuth, JWT and other related specs, which all seem to use timestamps in seconds.

For example, this was one of the trickest parts of trying to bring the legacy BrowserID assertion format into line with the final JWT spec.  It caused a pretty funny (and thankfully low-impact) security bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1096106

In the interests of consistency, I suggest we grit our teeth and use integer second timestamps throughout the public API.

lib/db/memory.js
```diff
@@ -57,6 +60,16 @@ const unique = require('../unique');
...
+ *       expiresAt: <timestamp>
+ *     }
+ *   },
+ *   refreshTokens: {
+ *     <token>: {
+ *       token: <string>,
+ *       clientId: <client_id>,
+ *       userId: <user_id>,
+ *       scope: <string>,
  *       createdAt: <timestamp>
```
rfk: Do you think it would be useful to track a `lastUsedAt` for refresh tokens?  We might want to e.g. expire tokens that have not been used in months.  Doesn't have to be part of the initial landing, but something to think about.

lib/db/memory.js
```diff
@@ -194,40 +210,67 @@ MemoryStore.prototype = {
...
   },
+  generateRefreshToken: function generateRefreshToken(vals) {
+    var token = unique.token();
+    var t = {
+      clientId: vals.clientId,
+      userId: vals.userId,
+      email: vals.email,
+      scope: vals.scope,
+      createdAt: new Date(),
+      token: encrypt.hash(token)
```
rfk: nit: it may be clearer to name this e.g. `tokenId` or `tokenHash` rather than just `token`

lib/db/mysql/index.js
```diff
@@ -148,13 +152,20 @@ const QUERY_CODE_INSERT =
...
   'VALUES (?, ?, ?, ?, ?, ?)';
 const QUERY_TOKEN_INSERT =
-  'INSERT INTO tokens (clientId, userId, email, scope, type, token) VALUES ' +
-  '(?, ?, ?, ?, ?, ?)';
+  'INSERT INTO tokens (clientId, userId, email, scope, type, expiresAt, ' +
+  'token) VALUES (?, ?, ?, ?, ?, ?, ?)';
+const QUERY_REFRESH_TOKEN_INSERT =
+  'INSERT INTO refreshTokens (clientId, userId, email, scope, token) VALUES ' +
+  '(?, ?, ?, ?, ?)';
 const QUERY_TOKEN_FIND = 'SELECT * FROM tokens WHERE token=?';
```
rfk: It's extra churn, but it may be worth renaming these to `QUERY_ACCESS_TOKEN_FIND` and similar in the interests of overall clarity.

lib/db/memory.js
rfk: Hmm...I see that would be inconsistent with the naming in the mysql backend, where we use `token` for the column name.  Probably not worth it then.

lib/scope.js
```diff
@@ -0,0 +1,55 @@
...
+function Scope(arr) {
+  if (!(this instanceof Scope)) {
+    return new Scope(arr);
+  }
+  if (!arr) {
+    arr = [];
+  } else if (arr instanceof Scope) {
+    return arr;
+  } else if (typeof arr === 'string') {
+    arr = arr.split(' ');
```
rfk: We could consider `split(/\s+/)` or similar here, to guard against weird whitespace in the input string.  Winding up with empty strings in the list of scopes seems like badness.

lib/db/mysql/index.js
```diff
@@ -378,10 +388,45 @@ MysqlStore.prototype = {
...
+      t.token = token;
+      return t;
+    });
+  },
+
+  getRefreshToken: function getRefreshToken(token) {
+    return this._readOne(QUERY_REFRESH_TOKEN_FIND, [buf(token)])
+    .then(function(t) {
+      if (t) {
+        t.scope = t.scope.split(' ');
```
rfk: Should this return a `Scope` instance here and in `getToken`?

lib/db/mysql/patches/patch-005-006.sql
```diff
@@ -0,0 +1,20 @@
...
+  token BINARY(32) PRIMARY KEY,
+  clientId BINARY(8) NOT NULL,
+  INDEX tokens_client_id(clientId),
+  FOREIGN KEY (clientId) REFERENCES clients(id) ON DELETE CASCADE,
+  userId BINARY(16) NOT NULL,
+  INDEX tokens_user_id(userId),
+  email VARCHAR(256) NOT NULL,
+  scope VARCHAR(256) NOT NULL,
+  createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+) ENGINE=InnoDB CHARACTER SET utf8 COLLATE utf8_unicode_ci;
```
rfk: I noticed that this table doesn't have a `type` column like the `tokens` table does.  Is it worth adding one?  It's not clear to me whether or not there could be different types of refresh token.

I guess we can always add one later if we introduce another token type.

lib/db/mysql/patches/patch-005-006.sql
```diff
@@ -0,0 +1,20 @@
...
+
+-- At expiresAt column for access tokens
+
+ALTER TABLE tokens ADD COLUMN expiresAt TIMESTAMP;
```
rfk: IIUC this will default to `NULL` for existing columns.  We should either to an `UPDATE tokens SET expiresAt =BLAH` as part of this migration, or ensure that the code checking these expiry times can handle `null` values.

lib/routes/token.js
```diff
@@ -11,21 +12,107 @@ const config = require('../config');
 const db = require('../db');
 const encrypt = require('../encrypt');
 const logger = require('../logging')('routes.token');
+const P = require('../promise');
+const Scope = require('../scope');
 const validators = require('../validators');
 
+const MAX_TTL = config.get('expiration.accessTokenS');
+const AUTHORIZATION_CODE = 'authorization_code';
+const REFRESH_TOKEN = 'refresh_token';
```
rfk: nit: suggest naming these `AUTHORIZATION_CODE_GRANT` or similar so it's a bit more obvious the context in which they'll be used.

lib/routes/token.js
```diff
@@ -11,21 +12,107 @@ const config = require('../config');
...
+      // + because loldatemath. without it, it does string concat
+      var expiresAt = +codeObj.createdAt + config.get('expiration.codeMs');
+      if (Date.now() > expiresAt) {
+        logger.debug('code.expired', { code: code });
+        throw AppError.expiredCode(code, expiresAt);
+      }
+    }
+    return codeObj;
+  });
+}
```
rfk: +1 to pulling these out into named helper functions, nice refactoring

lib/routes/token.js
```diff
@@ -11,21 +12,107 @@ const config = require('../config');
...
+
+function confirmToken(params) {
```
rfk: suggest `confirmRefreshToken` for clarity

lib/routes/token.js
```diff
@@ -11,21 +12,107 @@ const config = require('../config');
...
+        requested: params.scope
+      });
+      throw AppError.invalidScopes();
+    }
+    tokObj.scope = params.scope;
+    return tokObj;
+  });
+}
+
+function generateToken(options) {
```
rfk: suggest `generateTokens` plural to make it clear that multiple tokens might be returned

lib/routes/token.js
```diff
@@ -11,21 +12,107 @@ const config = require('../config');
...
+  return P.all(promises).spread(function(access, refresh) {
+    var json = {
+      access_token: access.token.toString('hex'),
+      token_type: access.type,
+      scope: Scope(access.scope).toString()
+    };
+    if (options.authAt) {
+      json.auth_at = options.authAt;
+    }
+    if (access.expiresAt) {
```
rfk: Under what circumstances could `access.expiresAt` be falsy?

lib/routes/token.js
```diff
@@ -33,10 +120,39 @@ var payloadSchema = Joi.object({
...
+    .valid(AUTHORIZATION_CODE, REFRESH_TOKEN)
+    .default(AUTHORIZATION_CODE)
+    .optional(),
+
+  ttl: Joi.number()
+    .max(MAX_TTL)
+    .default(MAX_TTL)
+    .optional(),
+
+  //XXX: forbidden??
```
rfk: Unresolved open question re: forbidden?

lib/routes/token.js
```diff
@@ -50,58 +166,37 @@ module.exports = {
...
-          if (Date.now() > expiresAt) {
-            logger.debug('code.expired', { code: code });
-            throw AppError.expiredCode(req.payload.code, expiresAt);
-          }
-        }
-        return code;
-      });
+      } else if (params.grant_type === REFRESH_TOKEN) {
+        return confirmToken(params).then(function(vals) {
+          vals.ttl = params.ttl;
```
rfk: I know it's unnecessary, but I'd consider putting an explicit `vals.generateRefreshToken = false` here for completeness.

lib/scope.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+  this._values = obj;
+}
+
+Scope.prototype = {
+
+  _values: undefined,
+
+  has: function has(scope) {
+    var that = this;
```
rfk: nit: I find `that` confusing as a name here, because it sounds like it's referring to some other object rather than the object in question; please consider `self` or `me` or similar.

lib/scope.js
```diff
@@ -0,0 +1,55 @@
...
+    return Scope(scope).values().every(function(word) {
+      if (word in that._values || word + ':write' in that._values) {
+        return true;
+      } else {
+        var prefix = word.split(':').slice(0, -1).join(':');
+        return prefix && that.has(prefix);
+      }
+    });
+  },
```
rfk: As an aside, we should formalize and document the precise rules we follow here so that reliers can follow convention.  We might also consider adding scope-checking to our verification endpoint that calls out to this logic.

lib/scope.js
```diff
@@ -0,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+function Scope(arr) {
```
rfk: again, nice refactor!

test/api.js
```diff
@@ -637,155 +639,330 @@ describe('/v1', function() {
...
+          };
+          db.registerClient(client2).then(function() {
+            mockAssertion().reply(200, VERIFY_GOOD);
+            return Server.api.post({
+              url: '/authorization',
+              payload: authParams({
+                client_id: client2.id.toString('hex')
+              })
+            }).then(function(res) {
+              console.log(res.result);
```
rfk: remove console.log in before merge

test/api.js
```diff
@@ -637,155 +639,330 @@ describe('/v1', function() {
...
+              }
+            });
+          }).then(function(res) {
+            assert.equal(res.statusCode, 200);
+            assert.equal(res.result.token_type, 'bearer');
+            assert(res.result.access_token);
+            assert.equal(res.result.access_token.length,
+              config.get('unique.token') * 2);
+            assert.equal(res.result.scope, 'foo bar');
+            assert.equal(res.result.auth_at, 123456);
```
rfk: should this also have returned a refresh token?

test/api.js
```diff
@@ -637,155 +639,330 @@ describe('/v1', function() {
...
             payload: {
               client_id: clientId,
               client_secret: secret,
-              code: url.parse(res.result.redirect, true).query.code,
-              foo: 'bar' // testing stripUnknown
+              grant_type: 'refresh_token',
+              refresh_token: unique.token().toString('hex')
             }
+          }).then(function(res) {
+            assert.equal(res.statusCode, 400);
```
rfk: Do we expect a particular errno or message?

test/api.js
```diff
@@ -637,155 +639,330 @@ describe('/v1', function() {
...
+              url: '/token',
+              payload: {
+                client_id: id2, // client2 stole it somehow
+                client_secret: secret2,
+                grant_type: 'refresh_token',
+                refresh_token: refresh
+              }
+            });
+          }).then(function(res) {
+            assert.equal(res.statusCode, 400);
```
rfk: ditto re: expected errno

test/scope.js
```diff
@@ -0,0 +1,68 @@
...
+      assert.deepEqual(s1.values(), ['a', 'b', 'c']);
+    });
+
+    it('should accept a Scope instance', function() {
+      var s1 = Scope(['a', 'b', 'c']);
+      var s2 = Scope(s1);
+      assert.equal(s1, s2);
+    });
+
+  });
```
rfk: Since the scope string comes from user input, I suggest adding some tests for weirdy values that might sneak through the Joi validator, like "foo         bar" and "profile::write:::".

test/scope.js
```diff
@@ -0,0 +1,68 @@
...
+      var s1 = Scope('foo bar');
+      assert(s1.has('foo'));
+      assert(s1.has('bar'));
+      assert(!s1.has('baz'));
+    });
+
+    it('should work with another Scope object', function() {
+      var s1 = Scope('foo bar');
+      var s2 = Scope('bar');
+      assert(s1.has(s2));
```
rfk: Worth checking that `!s2.has(s1)` because it's a subset, and you must have all scopes in the set.

test/scope.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    it('should allow sub-scopes', function() {
+      var s1 = Scope('foo bar:baz');
+      assert(s1.has('foo:dee'));
+      assert(s1.has('bar:baz'));
+      assert(s1.has('foo:mah:pa bar:baz:quux'));
```
rfk: Worth checking that it does *not* have `bar`.

rfk:
I didn't see any code (or tests) that check for access token expiry.  I guess the best place to do this is in the view, similar to how you currently check for code expiry?

rfk:
@seanmonstar this is pretty epic!  Lots of small comments in there but it's shaping up nicely.

IIUC there's no way to opt out of getting a refresh token in the authorization-code flow, right?  Existing clients will get both an access_token and a refresh_token, their access_token will expire in two weeks, and they will ignore the refresh token which will last forever.

I wonder if there's any way we can clean up these zombie refresh tokens, and whether it'll be enough of a problem in practice to bother.

lib/scope.js
seanmonstar: I'm saddened that `o[''] = 5` actually works, and is included in `Object.keys(o)`

rfk:
I wonder if it's worth linking the access_token back to the refresh_token that generated it?  Might be nice for bookkeeping purposes, and it would also allow a destroy of a refresh token to atomically revoke all access_tokens granted from it.

rfk:
We'll need to consider the behaviour of oauth-token-using code in Firefox with this API change; /cc @zaach, could you please comment when you've got a few moments?

The big client-facing chance here is that each `access_token` will now expire after (by default) two weeks.  The hope is that oauth clients will respond gracefully to the 401 from an expired token.  Firefox is particularly well situation here because it's capable of generating new tokens automatically.  But it's also the hardest thing for us to update if its existing behaviour would interact badly with this change.

We'll also have to reach out to web-based oauth consumers to at least make them aware of the change.  I'll gather a list to make sure we don't miss anyone.

rfk:
> I'll gather a list to make sure we don't miss anyone.

According to puppet-config, our active production reliers are:

* Firefox Marketplace
* Find My Device
* 123done.org
* 123done-prod.dev.lcip.org
* Loop/Hello
* Firefox implicit grant client
* Fennec implicit grant client
* FirefoxOS implicit grant client
* Firefox Accounts implicit grant client (is this the content-server itself?)
* Marketplace Payments
* Firefox Setup
* "FxAOAuthClient", whatever that is
* Pocket  web
* Pocket mobile

That's surprisingly many TBH.  But IIUC most of them should work without change because they don't use the access_token as a long-lived token, or should be easy to upgrade.

lib/db/mysql/patches/patch-005-006.sql
seanmonstar: Hm, perhaps this should update all access tokens to expire in 2 weeks (current max)?

lib/routes/token.js
seanmonstar: Originally, all access tokens generated before this PR lands. However, if we insert the date 2 weeks from now when altering the table, then all access tokens should have an expire date.

lib/scope.js
seanmonstar: The use of `that` is a common name in JavaScript, see suggestion by [Crockford](http://www.crockford.com/javascript/private.html)

lib/scope.js
seanmonstar: I'm not sure what you mean by calling this from the verification endpoint.

test/scope.js
seanmonstar: What's weird about `"foo bar"`?

test/scope.js
seanmonstar: added

test/scope.js
seanmonstar: added

test/api.js
seanmonstar: added

test/api.js
seanmonstar: added

lib/db/memory.js
seanmonstar: It probably would be useful, yes. Those that ignore the refresh token could be inadvertently generating new ones every time, and that would load up the table with a bunch of refresh tokens that sit there forever, ready to be guessed?

docs/api.md
seanmonstar: wuts grammer

seanmonstar:
@rfk updated!

seanmonstar:
@zaach ping (says rfk)

lib/scope.js
rfk: I mean that we could allow the SP to provide the list of expected scopes when verifying the token, and check it for them by calling out to this logic, e.g.:

```
> POST /v1/verify
> {
>   "token": "blahblahblah"
>   "scopes": "profile:email readinglist:write"
> }
...
< 40X Some Error Code
< {
<   message: "Scopes did not match"
< }
```

It might reduce footguns when implementing a new Service Provider.

test/scope.js
rfk: I...I don't know.  I suspect "foo bar" was a typo, but have no recollection of what I meant to type in its place.

docs/api.md
rfk: Suggest bolding **Seconds** here like you've done in the lists below

lib/scope.js
rfk: Still seems needlessly confusing to me, esp. in this context where we're operating on two instances of the same type...but I guess I can't argue with Crockford :-/

rfk:
It's still not clear to me where the actual expiry of access tokens is enforced.  If an RP shows up trying to use an access_token that has expired, what happens?

lib/scope.js
seanmonstar: Ok, I'll just completely remove the need for the binding. Instead, I'll pass `this` into `every()`, so that it's usable inside the closure,

seanmonstar:
Oh, great catch. `/verify` is supposed to check that the access token hasn't expired. I've added that in, with a corresponding test.

zaach:
Looking :eyes:

lib/config.js
```diff
@@ -80,10 +80,16 @@ const conf = convict({
     }
   },
   expiration: {
+    // in JavaScript, we live in milliseconds
+    accessToken: {
+      doc: 'Access Tokens maximum expiration (can live shorter)',
+      format: 'duration',
+      default: 1000 * 60 * 60 * 24 * 2 // 2weeks
```
zaach: convict will actually convert `'2 weeks'` into an integer.

lib/config.js
```diff
@@ -80,10 +80,16 @@ const conf = convict({
...
+    },
     code: {
       doc: 'Clients must trade codes for tokens before they expire',
       format: 'duration',
-      default: 1000 * 60 * 15
+      default: 1000 * 60 * 15 // 15mins
```
zaach: Same here: you can use `'15 minutes'`.

lib/config.js
seanmonstar: That's glorious! I'll fix this and add `strict` checking in a different PR.

rfk:
Fixes #209.

(Seeing if this will trick waffle into linking the two tickets...)

seanmonstar:
@zaach sanity check that it's fine we'll be expiring all tokens, even those made with an implicit grant (they'll expire in 2 weeks). Those using the implicit grant can just get a new token without bothering the user, and all others can grab the refresh_token and use that.

Both should already be handling a 400 from `/verify` for deactivated tokens, so receiving a 400 for expiration should be handled (they probably go through the login process again, but with refresh_tokens, that could be upgraded).

zaach:
Let me verify that desktop actual does handle 400s from `/verify`...

zaach:
Mark pointed out that we don't currently handle expiration, but it should not be too difficult to add: https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#385

rfk:
Fennec is also shipping oauth-token-generating code now, we'll need to check compatibility issues there as well:  https://bugzilla.mozilla.org/show_bug.cgi?id=1117829

rfk:
Just to add to what @zaach said, the current code in firefox doesn't handle expired or revoked tokens, but it's preffed off in all shipping versions.  So IIUC we should we able to tweak the behaviour for the next firefox release and not worry about breaking compat for older versions.  Which is excellent news :-)

rfk:
I've been sanity-checking the implications of putting this API in front of old oauth clients who don't understand refresh tokens.  The effects seem mostly harmless, as almost all of our reliers treat their access_token as single-use, and those that don't should really be dealing with revoked tokens already.

ISTM the only downside is we'll accumulate a bunch of zombie refresh tokens, thanks to clients receiving them but never doing anything with them.

To avoid this, we could consider having reliers explicitly opt-in to refresh tokens.  Note that I'm **not** suggesting that old reliers could continue to get long-lived access tokens.  But it might be nice to provide the choice between "just a short-lived access-token" and "a short-lived access-token plus the ongoing ability to generate more of them in future", since the former is what many of our reliers currently want.

@seanmonstar I'd love to hear your thoughts on whether this would make much difference in the system either way.

As prior art, the way G does this is fairly straightforward and documented at [1]; basically you just pass an additional `access_type` parameter in the authorization request like so:
```
https://accounts.google.com/o/oauth2/auth?
 scope=email%20profile&
 state=blahblah&
 response_type=code&
 client_id=foobar&
 access_type=offline
```

What's interesting is that this changes the UI that's presented to the user.  For `access_type=online` you're just asked if you want to share you information, but for `access_type=offline` you're asked if you give permission for them to access it *even when you're not online and using that service*.  We may be wise to consider such a distinction in more detail ourselves in future permission prompts.

Thoughts?

[1] https://developers.google.com/identity/protocols/OAuth2WebServer#offline

seanmonstar:
I'm fine with adding an `access_type=offline` parameter. Will do.

seanmonstar:
Hmm, implementing that would require storing the flag on the `code`. I'm less excited about that... but if it's only passed to the `/token` endpoint, there's no way to prompt the user if that's OK... Perhaps that doesn't matter, perhaps it does.

rfk:
Good point.  I think doing it via the code is the Right Thing in the overall architecture, but it does complicate the implementation.  Want to push it into a follow-up PR instead?

seanmonstar:
@rfk I feel like it's a breaking change to go from always providing a refresh_token to only giving one if you passed the correct parameter...

rfk:
*nod*; yeah I guess even if did them as two PRs we'd have to ship them on the same train.

What's your concern w.r.t managing this via the `codes` table?  IIRC we already do this with things like `authAt` so there's at least some precedent there.

seanmonstar:
I was just reluctant to alter the `codes` table. But I'll do it.

GitCop:

 * Commit: eabc9d1892444fa82b104ae7b6a3148bee06ac17

---

seanmonstar:
@rfk i added support for access_type=offline in a new commit, in case it's easier to review it separately. I'll squash them before merging.

GitCop:

 * Commit: 86886cdfc22669a4c2f09152dc893768eac87224

---

GitCop:

 * Commit: 5239ab41d72e24cda0ce4c708526f0326bf8ea0e

---

rfk:
Whoops, I think I accidentally commented on the commit rather than the diff; hopefully those two comments make sense.

This looks great @seanmonstar, thanks for separating the commits for review.  Please rebase when ready.

seanmonstar:
nits addressed, squashed into 1.

rfk:
I filed #275 as a follow-on for tracking last-used-at timestamps on refresh tokens.

rfk:
The access_type parameter will require support from the content-server; I filed https://github.com/mozilla/fxa-content-server/issues/2589

@seanmonstar I'm sure you'd be most welcome to dive in and implement that part of it as well, if you were feeling adventurous :-)

rfk:
> According to puppet-config, our active production reliers are:

I've dug into these, and the only one that I think we need to be careful with is pocket.  I'll send them an email and make sure we've got a clear path forward.

rfk:
(Migrating from IRC discussion)

Talking to Pocket, they'd like for all of their existing access_tokens to keep working after the switch, and I think this is a fair enough request.  We can achieve this by migating existing access tokens to become refresh tokens, staging the rollout like this:

1) Merge this code and get it out there and deployed for testing purposes
2) Pocket update their code, test, and start requesting refresh tokens
3) We do a db update to convert their previous access tokens into refresh tokens, perhaps using a timestamp-based filter to select only the ones that were created from a refresh token.

To give us enough time between (2) and (3) we could increase the default access_token lifetime to something longer, so we don't break things in the meantime.  Once reliers are migrated over we can drop it back down to a shorter lifetime.

@seanmonstar thoughts?

seanmonstar:
To upgrade Pocket's tokens:

```sql
INSERT INTO refreshTokens (token, clientId, userId, email, scope, createdAt)
  (SELECT token, clientId, userId, email, scope, createdAt FROM tokens WHERE clientId = '$POCKET_CLIENT_ID', createdAt < MIN(SELECT createdAt from refresh_tokens WHERE clientId = '$POCKET_CLIENT_ID'));
DELETE FROM tokens WHERE clientId = '$POCKET_CLIENT_ID';

```

rfk:
> To give us enough time between (2) and (3) we could increase the default access_token lifetime to something longer

@seanmonstar points out that this won't work, since expiry is based on token creation time.  But thinking about it some more, this won't matter anyway.  As long as pocket doesn't try to *use* the tokens between when they're created and when we switch them to refresh magic, nothing will break.

seanmonstar:
Alternatively, default expiration could be set to something ridiculous at first, and then re-adjusted afterwards.

seanmonstar:
rebased to master

rfk:
I think, based on IRC conversation, that we can do the above deployment plan on top of the code as it currently sits.  @seanmonstar can I get an official r? from you on that plan, and if so, I think this is ready to merge.

seanmonstar:
Yea that plan seems fine to me.

test/api.js
```diff
@@ -761,37 +835,187 @@ describe('/v1', function() {
...
+              payload: {
+                client_id: clientId,
+                client_secret: secret,
+                grant_type: 'refresh_token',
+                refresh_token: res.result.refresh_token,
+                ttl: MAX_TTL_S * 100
+              }
+            });
+          }).then(function(res) {
+            assertRequestParam(res.result, 'ttl');
```
rfk: Should this assert a particular response code?

rfk: n/m, I see the helper function does this internally.  That's not clear from the name, but I'll file a follow-up bug about it.

lib/routes/destroy.js
```diff
@@ -5,29 +5,38 @@
...
-    payload: {
-      token: Joi.string()
-        .length(config.get('unique.token') * 2) // hex = bytes*2
-        .regex(validators.HEX_STRING)
-        .required()
-    }
+    payload: Joi.object().keys({
+      access_token: validators.token,
+      refresh_token: validators.token
+    }).rename('token', 'access_token').xor('access_token', 'refresh_token')
```
rfk: I wonder if we should add a comment here about why it's being renamed, since we probably won't remember in a few month's time.

seanmonstar: Seems obvious to me, since token is now ambiguous, more specific names were needed.

rfk:
@seanmonstar my last-minute nits notwithstanding, I think we're solid here.  Let's land this as soon as train-40 has been cut, and maximise testing time in the wild.  r+ \o/

PR #262, opened 2015-05-29
WIP: Add token management endpoints

kitcambridge:
This adds two endpoints for managing the tokens associated with an account:

* `GET /user/tokens`: Gets all active OAuth tokens for an account. Each token record contains a `token_id` field (the token hash, but callers can treat this as opaque) that can be used to destroy the token.
* `DELETE /user/token/{id}`: Revokes a token associated with an account.

Both endpoints are authenticated with a bearer token containing a privileged `oauth:tokens` scope. The revocation endpoint could be used by the device manager service, which would store token hashes for each OAuth token granted by `FxAccountsOAuthClient`. When the user logs out remotely, the device manager would destroy all token hashes associated with that device.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -64,6 +64,9 @@ The currently-defined error responses are:
   - [DELETE /v1/client/:id][client-delete]
 - Developers
   - [POST /v1/developer/activate][developer-activate]
+- Users
+  - [GET /v1/user/tokens][user-tokens]
+  - [DELETE /v1/user/token/:id][user-token-delete]
 - [POST /v1/verify][verify]
 
 ### GET /v1/client/:id
@@ -469,6 +472,66 @@ A valid request will return JSON with these properties:
 }
 ```
 
+### GET /v1/user/tokens
+
+Gets a list of active tokens for a user. The `token_id` field of each token
+record is an opaque identifier that can be used to revoke the token.
+
+**Required scope:** `oauth:tokens`
+
+#### Request Parameters
+
+**Example:**
+
+```sh
+curl -v \
+-H "Authorization: Bearer 558f9980ad5a9c279beb52123653967342f702e84d3ab34c7f80427a6a37e2c0" \
+"https://oauth.accounts.firefox.com/v1/user/tokens"
+```
+
+#### Response
+
+A valid response will have a 200 status code and a list of token records:
+
+```json
+{
+  "tokens": [{
+    "token_id": "92784bc8b18985ae03a015bff9e6b6322e9a74a9673fa08bada418e4fb4e2f6a",
+    "client_id": "bc57cc3f2cfeb8c9",
+    "token_type": "bearer",
+    "scope": "profile:uid",
+    "created_at": 1432919962142
+  }, {
+    "token_id": "b0e7d5503130c72ba74ca238175f35609a65b8d441c7f09ac2cac43d469167f6",
+    "client_id": "77cae338e95db56c",
+    "token_type": "bearer",
+    "scope": "profile oauth:tokens",
+    "created_at": 1432920262142
+  }]
+}
+```
+
+### DELETE /v1/user/token/:id
+
+Revokes a token for an user.
+
+**Required scope:** `oauth:tokens`
+
+#### Request Parameters
+
+**Example:**
+
+```sh
+curl -v \
+-X DELETE \
+-H "Authorization: Bearer 558f9980ad5a9c279beb52123653967342f702e84d3ab34c7f80427a6a37e2c0" \
+"https://oauth.accounts.firefox.com/v1/user/token/92784bc8b18985ae03a015bff9e6b6322e9a74a9673fa08bada418e4fb4e2f6a"
+```
+
+#### Response
+
+A valid response will have a 204 response code and an empty body.
+
 [client]: #get-v1clientid
 [register]: #post-v1clientregister
 [clients]: #get-v1clients
@@ -480,3 +543,5 @@ A valid request will return JSON with these properties:
 [delete]: #post-v1destroy
 [verify]: #post-v1verify
 [developer-activate]: #post-v1developeractivate
+[user-tokens]: #get-v1usertokens
+[user-token-delete]: #delete-usertokenid
--- a/docs/api.md
+++ b/docs/api.md
@@ -460,6 +460,7 @@ A valid request will return JSON with these properties:
 - `client_id`: The client_id of the respective client.
 - `scope`: An array of scopes allowed for this token.
 - `email`: The email of the respective user.
+- `created_at`: The token creation time in milliseconds since Epoch.
 
 **Example:**
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -474,8 +474,8 @@ A valid request will return JSON with these properties:
 
 ### GET /v1/user/tokens
 
-Gets a list of active tokens for a user. The `token_id` field of each token
-record is an opaque identifier that can be used to revoke the token.
+Gets a list of active tokens for a user. The `id` field of each token record is
+an opaque identifier that can be used to revoke the token.
 
 **Required scope:** `oauth:tokens`
 
@@ -496,13 +496,13 @@ A valid response will have a 200 status code and a list of token records:
 ```json
 {
   "tokens": [{
-    "token_id": "92784bc8b18985ae03a015bff9e6b6322e9a74a9673fa08bada418e4fb4e2f6a",
+    "id": "92784bc8b18985ae03a015bff9e6b6322e9a74a9673fa08bada418e4fb4e2f6a",
     "client_id": "bc57cc3f2cfeb8c9",
     "token_type": "bearer",
     "scope": "profile:uid",
     "created_at": 1432919962142
   }, {
-    "token_id": "b0e7d5503130c72ba74ca238175f35609a65b8d441c7f09ac2cac43d469167f6",
+    "id": "b0e7d5503130c72ba74ca238175f35609a65b8d441c7f09ac2cac43d469167f6",
     "client_id": "77cae338e95db56c",
     "token_type": "bearer",
     "scope": "profile oauth:tokens",
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -16,6 +16,19 @@ exports.AUTH_STRATEGY = 'dogfood';
 exports.AUTH_SCHEME = 'bearer';
 
 exports.SCOPE_CLIENT_MANAGEMENT = 'oauth';
+exports.SCOPE_TOKEN_MANAGEMENT = 'oauth:tokens';
+
+function canManageClients(scopes) {
+  for (var i = 0, len = scopes.length; i < len; i++) {
+    var scope = scopes[i];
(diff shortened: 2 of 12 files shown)


seanmonstar:
Is the 'destroy' endpoint not satisfactory?

kitcambridge:
From the mailing list discussion, we decided to avoid having the device manager server store the original OAuth tokens. But the logic is essentially identical to `/destroy` otherwise.

rfk:
Right, this is for deleting a token if you only have its "id", as suggested in this thread: https://mail.mozilla.org/pipermail/dev-fxacct/2015-May/001514.html

@seanmonstar I'd love your thoughts on the suggested approach there overall.

We should consider this work in the context of #260, which is adding refresh-token support to the API.  We may want to hold off until after that lands, at which time this PR would have to revoke the refresh-token instead of (or as well as?) the access-token.

I also wonder if there's any value in pairing this with a "list all the tokens active on this account" API, which would be all the server-side support we need for a simple "view connected services and revoke the ones you don't like" dashboard.  That can be a separate PR, but it might inform the design of the API here.

@kitcambridge please update ./docs/api.md as well when you get a chance.

kitcambridge:
> We may want to hold off until after that lands, at which time this PR would have to revoke the refresh-token instead of (or as well as?) the access-token.

:+1:

> I also wonder if there's any value in pairing this with a "list all the tokens active on this account" API, which would be all the server-side support we need for a simple "view connected services and revoke the ones you don't like" dashboard. That can be a separate PR, but it might inform the design of the API here.

Having a dashboard like that would be lovely, especially as more services rely on FxA. And I see there's an API for developers to eventually register their own apps, too... :smiley:

Will add some explainer docs once I'm back at my laptop!

kitcambridge:
@seanmonstar, @rfk I went ahead and added an endpoint for listing tokens, since it seemed like an easier way to explain what the `DELETE` endpoint does, and where that ID comes from. Feedback (on the patch, and broader device manager proposal) welcome!

seanmonstar:
This seems like something the oauth-console could use as well, so ping @vladikoff.

vladikoff:
Yesss! listing tokens will be really useful for the console, thank you

lib/auth.js
```diff
@@ -16,6 +16,7 @@ exports.AUTH_STRATEGY = 'dogfood';
 exports.AUTH_SCHEME = 'bearer';
 
 exports.SCOPE_CLIENT_MANAGEMENT = 'oauth';
+exports.SCOPE_ACCOUNT_MANAGEMENT = 'account';
```
seanmonstar: Is there a reason to have a separate scope for these new endpoints? @vladikoff?

lib/routes/account/deleteToken.js
```diff
@@ -0,0 +1,41 @@
...
+
+module.exports = {
+  auth: {
+    strategy: auth.AUTH_STRATEGY,
+    scope: [auth.SCOPE_ACCOUNT_MANAGEMENT]
+  },
+  validate: {
+    params: {
+      token_id: Joi.string()
+      .length(encrypt.getHashSize() * 2) // hex = bytes*2
```
seanmonstar: indent 2 spaces

seanmonstar:
Generally looks good, just want to figure out the correct scopes here.

rfk:
In triage today we talked about trying to get this out on the next train, so setting the milestone appropriately

docs/api.md
```diff
@@ -465,6 +468,66 @@ A valid request will return JSON with these properties:
 }
 ```
 
+### GET /v1/account/tokens
+
+Gets a list of active tokens for an account. The `token_id` field of each token
```
seanmonstar: Just so I'm on the right page, "account" here is "user"?

lib/auth.js
seanmonstar: How about, since these scopes take up a global namespace of sorts, it be a sub-scope of `oauth`? So, to get tokens for a user, you'd need the scope `oauth:tokens` or something. The console, with its `oauth` scope, gets the subscope for free, and we leave "account" out of this. Thoughts @kitcambridge @rfk?

docs/api.md
kitcambridge: Yup, good catch! I'll change it to "user" for consistency.

lib/auth.js
kitcambridge: Sounds good to me. That's how `profile` works now, right?

rfk:
> In triage today we talked about trying to get this out on the next train,

This may of course depend on whether we ship the refresh-tokens stuff this train...

docs/api.md
```diff
@@ -465,6 +468,66 @@ A valid request will return JSON with these properties:
...
+```
+
+#### Response
+
+A valid response will have a 200 status code and a list of token records:
+
+```json
+{
+  "tokens": [{
+    "token_id": "92784bc8b18985ae03a015bff9e6b6322e9a74a9673fa08bada418e4fb4e2f6a",
```
rfk: I suggest "id" instead of "token_id" for symmetry with the existing `/clients` endpoint.

docs/api.md
```diff
@@ -465,6 +468,66 @@ A valid request will return JSON with these properties:
...
+    "created_at": 1432919962142
+  }, {
+    "token_id": "b0e7d5503130c72ba74ca238175f35609a65b8d441c7f09ac2cac43d469167f6",
+    "client_id": "77cae338e95db56c",
+    "token_type": "bearer",
+    "scope": "profile oauth:tokens",
+    "created_at": 1432920262142
+  }]
+}
+```
```
rfk: I'm sure I bring this up every time anyone implements a listing-style api, but we'll have to consider pagination here eventually.  We don't have to add it in this initial PR, but we should remember to address it before any on-device code to call this API ships to end users.

docs/api.md
```diff
@@ -465,6 +468,66 @@ A valid request will return JSON with these properties:
...
+
+### DELETE /v1/user/token/:id
+
+Revokes a token for an user.
+
+**Required scope:** `oauth:tokens`
```
rfk: Is it worth considering a read/write split with the permissions here, requiring `oauth:tokens:write` for deletion?  I can't think of any applications for it off the top of my head...

lib/auth.js
```diff
@@ -16,6 +16,19 @@ exports.AUTH_STRATEGY = 'dogfood';
...
+  for (var i = 0, len = scopes.length; i < len; i++) {
+    var scope = scopes[i];
+    if (scope === exports.SCOPE_CLIENT_MANAGEMENT ||
+      scope.lastIndexOf(exports.SCOPE_CLIENT_MANAGEMENT + ':', 0) === 0) {
+
+      return true;
+    }
+  }
+  return false;
+}
```
rfk: @seanmonstar's refresh-tokens PR has a nice refactor that will simplify this; https://github.com/mozilla/fxa-oauth-server/pull/260/files#r31217134

lib/routes/user/deleteToken.js
```diff
@@ -0,0 +1,41 @@
...
+const encrypt = require('../../encrypt');
+const unbuf = require('buf').unbuf.hex;
+const validators = require('../../validators');
+
+/*jshint camelcase: false*/
+
+module.exports = {
+  auth: {
+    strategy: auth.AUTH_STRATEGY,
+    scope: [auth.SCOPE_CLIENT_MANAGEMENT, auth.SCOPE_TOKEN_MANAGEMENT]
```
rfk: @seanmonstar, with your scope-checking cleanups in the other PR, would this reduce to just `auth.SCOPE_TOKEN_MANAGEMENT` and the broader scope would automatically imply the narrower?

lib/routes/user/deleteToken.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  },
+  handler: function deleteTokenId(req, reply) {
+    db.getToken(req.params.token_id)
+    .then(function(tok) {
+      if (!tok) {
+        throw AppError.invalidToken();
+      }
+      var token = tok.token;
+      return db.removeToken(unbuf(token));
```
rfk: It's not a big deal for the initial landing, but having to do separate "lookup" and "delete" queries for this is not awesome.  We should consider directly exposing a "delete token by id" operation as a single query.

lib/routes/user/deleteToken.js
```diff
@@ -0,0 +1,41 @@
...
+        .length(encrypt.getHashSize() * 2) // hex = bytes*2
+        .regex(validators.HEX_STRING)
+        .required()
+    }
+  },
+  handler: function deleteTokenId(req, reply) {
+    db.getToken(req.params.token_id)
+    .then(function(tok) {
+      if (!tok) {
+        throw AppError.invalidToken();
```
rfk: It's a tiny nit, but I'm not sure this is the right error here; to me "invalid token" is more of an authorization-level

test/api.js
```diff
@@ -1449,4 +1450,77 @@ describe('/v1', function() {
     });
   });
 
+  describe('user', function() {
+
+    var clientId = unique.id();
+    var tok, uid, vemail;
```
rfk: `vemail` appears unused?

rfk: actually so does `uid`

test/api.js
```diff
@@ -1449,4 +1450,77 @@ describe('/v1', function() {
...
+        headers: {
+          authorization: 'Bearer ' + tok
+        }
+      }).then(function(res) {
+        assert.equal(res.statusCode, 400);
+      });
+    });
+
+  });
+
```
rfk: We should make sure we add tests for failure cases here, e.g. trying to list tokens when you don't have the necessary scope.

rfk:
@kitcambridge this is looking great!  If you don't need this out in a rush, let's wait for the refresh-tokens stuff to land and then rebase on top of that.

lib/routes/user/deleteToken.js
seanmonstar: We have that. `db.removeToken(id)` does that. I imagine the lookup is so that the `invalidToken` error can be thrown. `removeToken` could probably be adjusted to return a boolean if the a row was deleted, allowing this to be one query...

lib/routes/user/deleteToken.js
seanmonstar: Hapi does that automatically. If `req.auth.credentials.scope` contains `oauth`, and this route demands `oauth:tokens`, Hapi will determine it as a match.

lib/routes/user/deleteToken.js
rfk: :+1:

GitCop:

 * Commit: e33b02fdf53b2a1f73a4faa3e0a49a5b845d1548

 * Commit: ed85b595077d2fb6a36bd40970744a2b35eb959a

---

GitCop:

 * Commit: e33b02fdf53b2a1f73a4faa3e0a49a5b845d1548

 * Commit: 69da2a206023d42799544d5ea6a4abe29942c6ff

---

rfk:
We moved away from this token-management approach for device manager, so I'm closing out this old PR.

PR #268, opened 2015-06-02
docs(verify): fix misnamed 'scopes' response property

### LINKED ISSUE #261: /verify endpoint documented as returning "scopes", actually returns "scope"
Per the documentation at:

  https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1verify

The verification endpoint should return an array of scopes under the name "scopes".  But the code actually returns it under the name "scope", causing confusion: https://github.com/mozilla/fxa-content-server/pull/2460/files#r31214570

IMHO "scopes" is better because it's an array, but we've probably got code in the wild that's checking for the value in "scope".  We'll either have to update the docs to match reality, or carefully migrate over to what the API says...

seanmonstar:
Closes #261

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -451,7 +451,7 @@ A valid request will return JSON with these properties:
 
 - `user`: The uid of the respective user.
 - `client_id`: The client_id of the respective client.
-- `scopes`: An array of scopes allowed for this token.
+- `scope`: An array of scopes allowed for this token.
 - `email`: The email of the respective user.
 
 **Example:**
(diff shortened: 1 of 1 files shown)


vladikoff:
+1 via https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/verify.js

```
 response: {
    schema: {
      user: Joi.string().required(),
      client_id: Joi.string().required(),
      scope: Joi.array(),
      email: Joi.string()
    }
  },
```

PR #273, opened 2015-06-10
chore(config): Update convict and switch on strict validation.

### LINKED ISSUE #263: set convict to strict validation
This will catch when any config value is added that shouldn't have been, or typos. Requires a bump to convict@0.8

rfk:
Fixes #263

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -196,7 +196,11 @@ var files = (envConfig + ',' + process.env.CONFIG_FILES)
   .split(',').filter(fs.existsSync);
 conf.loadFile(files);
 
-conf.validate();
+var options = {
+  strict: true
+};
+
+conf.validate(options);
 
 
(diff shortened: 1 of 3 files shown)


rfk:
dependency-related travis failures....*shrug*

rfk:
@chilts r?

rfk:
Note that I've disabled the check-for-vulnerable-modules thing since it was failing due to an awsbox dependency, and we don't have an upgrade path for that.  IIRC we disabled that check in other repos anyway, so it seems like the simplest way to remain unblocked here.

rfk:
Actually I take that back - it seems like the approach in this repo is to not include dev dependencies in the shrinkwrap file, which removes the reported vulnerable module.

chilts:
Tests pass locally, Travis is happy, code lgtm. r+

Issue #194, opened 2015-01-09
Fail fast on invalid `action` parameters to GET /v1/authorization


shane-tomlinson:
From @rfk in https://github.com/mozilla/fxa-oauth-server/pull/191/files#r22686284:

> Should we consider erroring out if an invalid action is specified? I can imagine people accidentally typoing this as e.g. force_signin.

I agree.

rishibaldawa:
This file https://github.com/mozilla/fxa-oauth-server/blob/07efe68605c78b31c3babffeeb398530c9bb66cf/lib/routes/redirect.js#L18 ?

rfk:
Yep, that's the one!

rishibaldawa:
What if no action is provided? There's a test for this that'll fail if we just throw a invalidRedirect https://github.com/mozilla/fxa-oauth-server/blob/master/test/api.js#L224

Do we need to fix the test or continue the behavior ?

seanmonstar:
The test is correct. No action should mean the default.

PR #277, opened 2015-06-19
fix(api): fail on invalid action parameters

rishibaldawa:
fixes https://github.com/mozilla/fxa-oauth-server/issues/194

### DIFF
--- a/lib/routes/redirect.js
+++ b/lib/routes/redirect.js
@@ -5,6 +5,7 @@
 const url = require('url');
 
 const config = require('../config');
+const AppError = require('../error');
 
 function actionToPathname(action) {
   if (action === 'signup') {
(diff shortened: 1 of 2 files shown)


lib/routes/redirect.js
```diff
@@ -13,9 +14,11 @@ function actionToPathname(action) {
     return 'force_auth';
   } else if (action === 'signin') {
     return 'signin';
+  } else if (action === undefined) {
+    return '';
   }
 
-  return '';
+  throw AppError.incorrectRedirect();
```
seanmonstar: This error actually is for when the `redirect_uri` doesn't match with what was registered for the client.

seanmonstar:
This could be handled by putting Joi validation on the `query` part of the route... Allowing unknown, and having a single rule for `action` with `Joi.valid('signin', 'signup', 'force_auth', '')` would do it. Or, just use the `AppError.invalidRequestParameters('action')`.

rishibaldawa:
Using Joi validation will require a whole of boilerplate and hence maintenance. If there's more reasons to have a list/array of actions in one place then it would be a good choice but just for validations it feels too much work. I'll go with `AppError.invalidRequestParameters('action')` for now but lemme know if you prefer the former.

PR #278, opened 2015-06-23
feat(log): Add activity event logging

chilts:


chilts:
WIP ... :)

chilts:
Added three events for generateCode, generateGrant and token. Not sure if we need more or less than this. Also logs the `client_id` if it exists.

I've added the same logging of user-agent, though in the case of token I guess that just logs the server user-agent ... which in my testing was empty (from 123done).

Using `fxa-local-dev`, here's an example of each (plus some line breaks):

```
oauth-server PORT 9010-6 (out): fxa-oauth-server.activity.INFO:
  activityEvent
  {"event":"generateGrant","uid":"cf329a5e8bc3417eac215ac678bfbba6","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:40.0) Gecko/20100101 Firefox/40.0","client_id":"dcdb5ae7add825d2"}

oauth-server PORT 9010-6 (out): fxa-oauth-server.activity.INFO:
  activityEvent
  {"event":"generateGrant","uid":"91b76000a11e44b5abc90eded49a92eb","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:40.0) Gecko/20100101 Firefox/40.0","client_id":"98e6508e88680e1a"}

oauth-server PORT 9010-6 (out): fxa-oauth-server.activity.INFO:
  activityEvent
  {"event":"token","uid":"9368396f361e4613ab0845ded7dc1f8a","client_id":"dcdb5ae7add825d2"}
```

chilts:
The way the request is being passed to `toToken` so we can call `activityEvent` with it could be a little nicer I think.

rfk:
> Added three events for generateCode, generateGrant and token

Verifying a token is also important for MAU tracking purposes.

chilts:
Thanks @rfk.

chilts:
So it looks like I can get the `client_id` from the `token` in verify ... but not the request as I have currently assumed. I'll do a bit of re-arranging so this makes more sense instead of passing req in to the `activityEvent()` function.

chilts:
This should be a bit simpler now since the client id is being pulled from the relevant token where available. Ready for review.

lib/logging/activity-event.js
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const logger = require('./')('activity');
+
+module.exports = function activityEvent(event, uid, client_id, userAgent) {
```
rfk: I suggest taking the full `request` object as argument here and pulling the user-agent out of it, in case we decide to log more details in future.  This more closely parallels what is done in the auth-server and it can still be `null` for server-to-server requests.

lib/routes/authorization.js
```diff
@@ -80,17 +81,19 @@ function generateCode(claims, client, scope, req) {
...
 
-function generateGrant(claims, client, scope) {
+function generateGrant(claims, client, scope, req) {
   return db.generateToken({
     clientId: client.id,
     userId: buf(claims.uid),
     email: claims['fxa-verifiedEmail'],
     scope: scope
   }).then(function(token) {
+    activityEvent('generateGrant', claims.uid, hex(client.id), req.headers['user-agent']);
```
rfk: I'm not sure there's user-facing or metrics-facing value in having separate `generateGrant` and `generateCode` events.  The high-level action that took place here was "authorized a relier", please consider combining these into a single `authorize` event that can be logged further up the chain where you already have the `request` object available.

lib/routes/token.js
```diff
@@ -19,6 +20,7 @@ function generateToken(code) {
 }
 
 function toToken(authAt, _, token) {
+  activityEvent('token', hex(token.userId), hex(token.clientId), null /* server-to-server */);
```
rfk: Likewise, at a high level this event is part of the code-based `authorize` flow.  Do we need it called out separately?

lib/routes/verify.js
```diff
@@ -26,6 +27,10 @@ module.exports = {
     }
   },
   handler: function verify(req, reply) {
-    token.verify(req.payload.token).done(reply, reply);
+    token.verify(req.payload.token).done(
+      function(tokenInfo) {
+        activityEvent('verify', tokenInfo.user, tokenInfo.client_id, null /* server-to-server */);
```
rfk: :+1:

lib/logging/activity-event.js
```diff
@@ -0,0 +1,19 @@
...
+  var info = {
+    event: event,
+    uid: uid,
+    client_id: client_id,
+  };
+  // we're called from either a browser or a server, so check if we have this
+  if ( userAgent ) {
+    info.userAgent = userAgent;
+  }
+  logger.info('activityEvent', info);
```
rfk: Will this come out in the same structure as the events from the auth-server?

lib/logging/activity-event.js
chilts: Good idea.

lib/logging/activity-event.js
chilts: Not sure at the moment, am just looking. Will paste examples into the issue as a comparison.

lib/routes/authorization.js
chilts: Ok. Am just doing a rebase due to the changes in https://github.com/mozilla/fxa-oauth-server/pull/260 so a few things have moved.

From what I can tell, if I move it higher up, I don't yet have the 'claims' which I'm currently getting the uid from. I'll rebase and continue from there, there might be another way.

lib/routes/token.js
chilts: So we don't need this at all?

rfk: right, I don't think we need it as a separate event

chilts:
Some of the events look like this (when using fxa-local-dev):

```
oauth-server PORT 9010-6 (out): fxa-oauth-server.activity.INFO: activityEvent {"event":"authorizedRelier","uid":"0b89eaf9b9384a558363aca83a4b4d69","client_id":"dcdb5ae7add825d2","userAgent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:38.0) Gecko/20100101 Firefox/38.0"}
oauth-server PORT 9010-6 (out): fxa-oauth-server.activity.INFO: activityEvent {"event":"authorization","uid":"0b89eaf9b9384a558363aca83a4b4d69","client_id":"dcdb5ae7add825d2"}
oauth-server PORT 9010-6 (out): fxa-oauth-server.activity.INFO: activityEvent {"event":"verify","uid":"0b89eaf9b9384a558363aca83a4b4d69","client_id":"dcdb5ae7add825d2"}
```

The only event I couldn't trigger was the `refreshToken` in the `tokenEndpoint`. @rfk any thoughts on how I can trigger that one?

rfk:
We're going to revisit this stuff as part of some broader metrics-related user stories, closing this out for now.

Issue #159, opened 2014-10-20
improve `grunt release` to bump version


seanmonstar:

rishibaldawa:
Something like [grunt-bump](https://github.com/geddski/grunt-release) or [grunt-release](https://github.com/geddski/grunt-release) ?

seanmonstar:
Oh nice, grunt-release looks perfect.

rishibaldawa:
I have some time this weekend. Will try to get something working.

pdehaan:
FWIW, we're using **grunt-bump** for the [**fxa-content-server**](https://github.com/mozilla/fxa-content-server/blob/7c38abe5093c5c43b1d2d1c29bade2c79f542730/package.json#L78) and [**fxa-auth-server**](https://github.com/mozilla/fxa-auth-server/blob/master/package.json#L51) repos. /cc @jrgm

rishibaldawa:
using the same thing everywhere makes much more sense to me, so I'd prefer to start with grunt-bump. If grunt-bump doesn't suit all usecases, then migrating all to grunt-release later would be easier. Let me know what you think.

rfk:
:+1: to consistency

PR #288, opened 2015-07-01
chore(release): add tasks "grunt version" and "grunt version:patch" 

rishibaldawa:
to create release tags

Fixes https://github.com/mozilla/fxa-oauth-server/issues/159

### DIFF
--- a/tasks/bump.js
+++ b/tasks/bump.js
@@ -1,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// takes care of bumping the version number in package.json
+
+module.exports = function (grunt) {
+  'use strict';
+
+  grunt.config('bump', {
+    options: {
+      files: ['package.json', 'npm-shrinkwrap.json'],
+      bumpVersion: true,
+      commit: true,
+      commitMessage: 'Release v%VERSION%',
+      commitFiles: ['package.json', 'npm-shrinkwrap.json', 'CHANGELOG'],
+      createTag: true,
+      tagName: 'v%VERSION%',
+      tagMessage: 'Version %VERSION%',
+      push: false,
+      pushTo: 'origin',
+      gitDescribeOptions: '--tags --always --abrev=1 --dirty=-d'
+    }
+  });
+};
--- a/tasks/bump.js
+++ b/tasks/bump.js
@@ -13,7 +13,7 @@ module.exports = function (grunt) {
       bumpVersion: true,
       commit: true,
       commitMessage: 'Release v%VERSION%',
-      commitFiles: ['package.json', 'npm-shrinkwrap.json', 'CHANGELOG'],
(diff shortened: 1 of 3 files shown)


tasks/bump.js
```diff
@@ -0,0 +1,25 @@
...
+    options: {
+      files: ['package.json', 'npm-shrinkwrap.json'],
+      bumpVersion: true,
+      commit: true,
+      commitMessage: 'Release v%VERSION%',
+      commitFiles: ['package.json', 'npm-shrinkwrap.json', 'CHANGELOG'],
+      createTag: true,
+      tagName: 'v%VERSION%',
+      tagMessage: 'Version %VERSION%',
+      push: false,
```
jrgm: Thanks, @rishibaldawa 

I'll note one thing about this change. At present, the tags have been of the form 'L.M.N', which is different from fxa-{content,auth,customs,auth-db,etc.}-server' repositories tagging. I'd been meaning to ask @seanmonstar if he was okay with that change. But the inconsistency of the tagging pattern has been something that is just a nuisance in deployments.

Otherwise, LGTM!

rfk: FWIW, I think we should strongly favour consistency with the other repos in our cluster

seanmonstar: I was just following orders from @ckolos! *peon voice* werk werk!

jrgm: So, the `npm version` command, by default, will create tags with a leading `v`. @ckolos - do you have a particular reason to want '1.2.3' with no leading `v` in the tag names?

rishibaldawa: I'd favor consistency as well. Either works though `fxa-oauth-server v1.2.3` is much better than `fxa-oauth-server 1.2.3` in explaining that `1.2.3` is the version. If we go with not having the `v` prefix, I'd recommend not having it in rest of the packages as well.

rfk: I nominate @jrgm to pick the format he likes best and just enforce it uniformly everywhere :-)

jrgm:
V for victory!

tasks/bump.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = function (grunt) {
+  'use strict';
+
+  grunt.config('bump', {
+    options: {
+      files: ['package.json', 'npm-shrinkwrap.json'],
+      bumpVersion: true,
+      commit: true,
+      commitMessage: 'Release v%VERSION%',
+      commitFiles: ['package.json', 'npm-shrinkwrap.json', 'CHANGELOG'],
```
jrgm: @rishibaldawa - So in this repository, CHANGELOG is CHANGELOG.md. It's not worth it to change that name in this repo, so can you adjust the name in the line above?

rishibaldawa: sounds good.

rishibaldawa:
Charge!!!!

jrgm:
So works for me. 

(As a side note for future reference, a couple of things: 1. it is better for you to create a feature branch in your repo for a pull request (instead of from `master` in your repo, and 2. it's also better to squash the little fix-ups to this PR and force push that to your repo; makes git history cleaner, but I'm not going to worry about it here).

PR #298, opened 2015-07-16
feat(api): add ttl parameter to POST /authorization

### LINKED ISSUE #292: Allow "ttl" property for implicit token grants
When using the implicit flow to create an access token, it should be possible to pass "ttl" paramter to customize the expiry time of the token.

ref https://bugzilla.mozilla.org/show_bug.cgi?id=1181959

philbooth:
Fixes #292.

The new parameter is optional when `response_type=token` and forbidden when `response_type=code`.

Assumptions:

* It is expressed in seconds.
* The maximum should be `config.get(expiration.accessToken)`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -318,6 +318,7 @@ back to the client. This code will be traded for a token at the
 - `assertion`: A FxA assertion for the signed-in user.
 - `state`: A value that will be returned to the client as-is upon redirection, so that clients can verify the redirect is authentic.
 - `response_type`: Optional. If supplied, must be either `code` or `token`. `code` is the default. `token` means the implicit grant is desired, and requires that the client have special permission to do so.
+- `ttl`: Optional if `response_type=token`, forbidden if `response_type=code`. Indicates the requested lifespan in seconds for the implicit grant token. The value is subject to an internal maximum limit, so clients must check the `expires_in` result property for the actual TTL.
 - `redirect_uri`: Optional. If supplied, a string URL of where to redirect afterwards. Must match URL from registration.
 - `scope`: Optional. A string-separated list of scopes that the user has authorized. This could be pruned by the user at the confirmation dialog.
 - `access_type`: Optional. A value of `offline` will generate a refresh token along with the access token.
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -318,6 +318,7 @@ back to the client. This code will be traded for a token at the
 - `assertion`: A FxA assertion for the signed-in user.
 - `state`: A value that will be returned to the client as-is upon redirection, so that clients can verify the redirect is authentic.
 - `response_type`: Optional. If supplied, must be either `code` or `token`. `code` is the default. `token` means the implicit grant is desired, and requires that the client have special permission to do so.
+- `ttl`: Optional if `response_type=token`, forbidden if `response_type=code`. Indicates the lifespan in seconds for the requested implicit grant token. If supplied, it must be a positive number less than or equal to 1,209,600 (2 weeks).
```
rfk: I expect we will reduce this significantly in future, so maybe not worth calling it out in the docs.  Rather we can say that the server may cap this to an internal maximum value, and clients should check the `expiresIn` return value for the actual TTL.

lib/routes/authorization.js
```diff
@@ -76,12 +77,13 @@ function generateCode(claims, client, scope, req) {
   });
 }
 
-function generateGrant(claims, client, scope) {
+function generateGrant(claims, client, scope, request) {
```
rfk: please name this param `req` for consistency with the `generateCode` function above.  Or rename the one in `generateCode` to match, if you feel strongly about it :-)

test/api.js
```diff
@@ -566,14 +578,32 @@ describe('/v1', function() {
...
             })
           }).then(function(res) {
+            var defaultExpiresIn = config.get('expiration.accessToken') / 1000;
             assert.equal(res.statusCode, 200);
             assert(res.result.access_token);
             assert.equal(res.result.token_type, 'bearer');
             assert(res.result.scope);
-            assert(res.result.expires_in);
+            assert(res.result.expires_in <= defaultExpiresIn);
+            assert(res.result.expires_in > defaultExpiresIn - 10);
```
rfk: :+1:

rfk:
> Assumptions:
>    It is expressed in seconds.
>    The maximum should be config.get(expiration.accessToken).

Yes, good call on both counts.  Thanks @philbooth, r+ with a couple of small nits.

lib/routes/authorization.js
philbooth: Ha, I don't feel *that* strongly about it. :)

philbooth:
Merged, with the suggested changes.

PR #303, opened 2015-07-19
fix(sql): remove references to the `whitelisted` column; this is now the `trusted` column

### LINKED ISSUE #301: Commit 155d2ce (to drop `whitelisted` and rename `secret` to `hashedSecret`) cannot go to production
re: https://github.com/mozilla/fxa-oauth-server/pull/284 and https://github.com/mozilla/fxa-oauth-server/commit/155d2ce39967513382e38948c9490a327d4441e8

I cannot do:
```
ALTER TABLE clients CHANGE secret hashedSecret BINARY(32);
ALTER TABLE clients DROP COLUMN whitelisted;
```
to prepare the database for oauth train-42. Either one will cause an immediate production outage.

Database changes must always be compatible with train-N (production) and train-(N+1) (the next one). 

Column renames can never be done. Dropping a column can only happen AFTER the current production stable release no longer refers at all to the column that will be dropped.

I have two pull requests coming up to fix this for train-42:

1. Revert commit `155d2ce`, but keeping the patches change and adding to that changes to reverse the two ALTER statements. (This is to avoid breaking `latest` and other, and having to manually fix things). 

2. A follow-on PR to (1) which puts back in the changes to remove all references to `whitelisted`, which is the second step in obsoleting this column (the third step is actually dropping the column AFTER that has gone live in production).

/cc @ckolos

jrgm:
fixes #301 

But this should only be merged after GH-302 

Puts back the parts that remove references to a `whitelisted` column, which can be dropped AFTER the train with this change is live and stable.

r? @seanmonstar

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -113,8 +113,8 @@ MysqlStore.connect = function mysqlConnect(options) {
 const QUERY_CLIENT_REGISTER =
   'INSERT INTO clients ' +
   '(id, name, imageUri, secret, redirectUri, termsUri, privacyUri, ' +
-  ' whitelisted, trusted, canGrant) ' +
-  'VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?);';
+  ' trusted, canGrant) ' +
+  'VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);';
 const QUERY_CLIENT_DEVELOPER_INSERT =
   'INSERT INTO clientDevelopers ' +
   '(rowId, developerId, clientId) ' +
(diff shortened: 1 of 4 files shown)


jrgm:
I dunno. I may need to rebase this after #302 is merged.

rfk:
@jrgm ok, the dependency PR is merged, feel free to rebase and merge as appropriate. I'm r+ on the second commit that drops "whitelisted", assuming that it's `DEFAULT FALSE` in the production schema.

jrgm:
Rebased and merged (although maybe I didn't need a rebase, I did it anyways).

jrgm:
Sorry forgot  to address your review comment:

> I'm r+ on the second commit that drops "whitelisted", assuming that it's DEFAULT FALSE in the production schema.

Yes, that column, whitelisted, is default false.

PR #305, opened 2015-07-19
fix(server): exit if db patch level is wrong

### LINKED ISSUE #291: oauth server hangs if database version number is not the expected one
I've bumped into this in the past. Maybe there is already an open bug about this. Anyways, if there is a problem with the db config, the process starts up but cannot connect to the db and just sits there with only these log messages;
```
{"Timestamp":1436411814121000000,"Logger":"fxa-oauth-server","Type":"db.mysql.checkDbPatchLevel","Severity":2,"Pid":2337,"EnvVersion":"2.0","Fields":{"msg":"unexpected db patch level: 5"}}                                                                                           
{"Timestamp":1436411814122000000,"Logger":"fxa-oauth-server","Type":"db.ping","Severity":2,"Pid":2337,"EnvVersion":"2.0","Fields":{}}       
{"Timestamp":1436411814122000000,"Logger":"fxa-oauth-server","Type":"bin.server.db.ping","Severity":0,"Pid":2337,"EnvVersion":"2.0","Field":{}}
```

Just dying would be better behaviour.

philbooth:
Fixes #291.

Although the fix itself was straightforward, adding tests for it was less so. There were no tests in place for `lib/db/mysql/index.js` and I didn't want to make a change without adding appropriate tests. To that end, I've introduced two new dev dependencies, `sinon` and `mockery`.

You'll notice some fairly ugly boilerplate at the top of the new test module, `test/db/mysql.js`. This is partly because there are a lot of dependencies to mock but also because the blanket coverage tool rewrites `require` paths in the instrumented code, so we need to do a bit of magic to ensure mockery is handling them correctly.

You may or may not be thinking that this seems like a lot of hoops to jump through, just for testing a new call to `process.exit()`. I think they're worth it though, because they pave the way for easily adding further tests for this module down the line.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,9 @@
     "load-grunt-tasks": "^3.1.0",
     "mocha-text-cov": "^0.1.0",
     "nock": "^1.2.1",
+    "proxyquire": "^1.6.0",
     "read": "^1.0.5",
+    "sinon": "^1.15.4",
     "time-grunt": "^1.1.0"
   }
 }
--- a/test/lib/mocks.js
+++ b/test/lib/mocks.js
@@ -1,0 +1,111 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const path = require('path');
+const proxyquire = require('proxyquire');
+const m = require('module');
+
+var moduleCache;
+
+module.exports = {
+  require: requireDependencies,
+  register: registerDependencies,
+  deregister: deregisterDependencies
+};
+
+// `mocks.require`
+//
+// Require dependencies using the same path that is specifed in the module
+// under test.
+//
+// Returns an object containing dependencies keyed by their path.
+//
+// Expects three arguments; `dependencies`, `modulePath` and `basePath`.
+//
+// dependencies: Array of { path, ctor } items.
+//               path: The dependency path, as specified in the module
+//                     under test.
+//               ctor: Optional. If the dependency is a constructor for
+//                     an instance that you wish to mock, set this to a
+//                     function that returns your mock instance.
+// modulePath:   The relative path to the module under test.
+// basePath:     The base path, i.e. __dirname for the test itself.
+function requireDependencies(dependencies, modulePath, basePath) {
+  var result = {};
+
+  dependencies.forEach(function (dependency) {
+    result[dependency.path] = requireDependency(dependency, modulePath, basePath);
+  });
+
+  return result;
+}
+
+function requireDependency(dependency, modulePath, basePath) {
+  if (typeof dependency.ctor === 'function') {
+    return dependency.ctor;
+  }
+
+  var localPath = dependency.path;
+
+  if (localPath[0] === '.') {
+    localPath = path.relative(
+      basePath,
+      path.resolve(basePath, modulePath, localPath)
+    );
+  }
+
+  return require(localPath);
+}
+
+// `mocks.register`
+//
+// Register mock dependencies, fixing paths as we go so that it works
+// with the blanket coverage tool (which rewrites require paths in the
+// instrumented code). You should call this function inside beforeEach.
+//
+// Expects three arguments; `dependencies`, `modulePath` and `basePath`.
+//
+// dependencies: An object, where keys are dependency paths and values
+//               are mock objects. This argument is typically the return
+//               value from `mocks.require`, modified by sinon for your
+//               tests.
+// modulePath:   The relative path to the module under test.
+// basePath:     The base path, i.e. __dirname for the test itself.
+function registerDependencies(dependencies, modulePath, basePath) {
+  var instrumentedDependencies = {};
+
+  clearModuleCache();
+
+  Object.keys(dependencies).forEach(function(dependencyPath) {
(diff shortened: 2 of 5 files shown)


philbooth:
I'm a bit confused about the failing build. The tests run fine for me locally.

This is the error:

```
fxa-oauth-server.db.mysql.ERROR: checkDbPatchLevel Error: Cannot find module 'isarray'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/home/travis/build/mozilla/fxa-oauth-server/node_modules/mysql/node_modules/readable-stream/lib/_stream_readable.js:25:15)
    at Module._compile (module.js:456:26)
    at Module._extensions..js (module.js:474:10)
    at Object.require.extensions..js (/home/travis/build/mozilla/fxa-oauth-server/node_modules/blanket/src/index.js:179:17)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/home/travis/build/mozilla/fxa-oauth-server/node_modules/mysql/node_modules/readable-stream/readable.js:1:90)
    at Module._compile (module.js:456:26)
    at Module._extensions..js (module.js:474:10)
    at Object.require.extensions..js (/home/travis/build/mozilla/fxa-oauth-server/node_modules/blanket/src/index.js:179:17)
```

`isarray` is only mentioned in `npm-shrinkwrap.json`, as a dependency of `readable-stream`. Any ideas what I've done wrong?

jrgm:
@philbooth - this is almost certainly npm silently failing its install. I noticed the same with my PR's last night. Just re-running travis gets this green. I put in pull requests to fxa-oauth and fxa-profile to use npm@2. (I have a script to fix up other repos, but I have to head out now for a while, and will fix up the others later today).

philbooth:
Thanks @jrgm! :)

rfk:
> To that end, I've introduced two new dev dependencies, sinon and mockery.

+1 to `sinon`, I've enjoyed using it in my forays into the frontend code.

> You'll notice some fairly ugly boilerplate at the top of the new test module,

It looks like most of this is not terribly specific to the mysql tests, is it worth factoring it out into a utility module "mocks.js" or similar, so that we're encouraged to use it for additional tests in the future?

rfk:
Also /cc @dannycoates for an additional perspective on mockery.  I know we jump through some hoops in fxa-auth-server so that we can directly pass in mocked dependencies rather than fiddling with `require()`, any thoughts on the relative merits of taking that approach vs something like the use of mockery here?

package.json
```diff
@@ -55,6 +55,8 @@
     "mocha-text-cov": "^0.1.0",
     "nock": "^1.2.1",
     "read": "^1.0.5",
-    "time-grunt": "^1.1.0"
+    "time-grunt": "^1.1.0",
+    "mockery": "^1.4.0",
+    "sinon": "^1.15.4"
```
jrgm: Small nit, but looks previous authors have tried to keep the list in alphabetical order.

philbooth:
I've pushed changes for the feedback received so far, thanks for your comments.

The new mocking helper functions are in `test/util/mocks.js`. I also took the liberty of moving `test/lib/server.js` into the same directory. I wanted to limit the number of directories that we exclude from mocha, but it felt wrong to have `mocks.js` inside a directory called `lib`. But I'm happy to revert that if it doesn't sit well with you guys.

philbooth:
> The new mocking helper functions are in `test/util/mocks.js`. I also took the liberty of moving `test/lib/server.js` into the same directory. I wanted to limit the number of directories that we exclude from mocha, but it felt wrong to have `mocks.js` inside a directory called `lib`.

Changed my mind about this. :smile:

The new mocking helper functions are now in `test/lib/mocks.js` and `test/lib/server.js` remains at the same path.

philbooth:
Over on the content server, Vlad's asked me to try using [proxyquire](https://github.com/thlorenz/proxyquire) instead of mockery because it's a more active project. Assuming that goes well, I'll come back and make the switch here too.

Just mentioning in case anyone considers merging this PR: **DON'T MERGE IT YET**:exclamation:

philbooth:
Rebased with `proxyquire` replacing `mockery`.

tasks/test.js
```diff
@@ -14,7 +14,7 @@ module.exports = function (grunt) {
       },
       src: [
         'test/**/*.js',
-        '!test/lib/**'
+        '!test/util/**'
```
rfk: Looks like you need to revert this after switching back to from `util` to `lib`?

test/lib/mocks.js
```diff
@@ -0,0 +1,111 @@
...
+
+  Object.keys(dependencies).forEach(function(dependencyPath) {
+    var instrumentedPath = getInstrumentedPath(dependencyPath, modulePath, basePath);
+    instrumentedDependencies[instrumentedPath] = dependencies[dependencyPath];
+  });
+
+  proxyquire(modulePath, instrumentedDependencies);
+}
+
+function clearCache() {
```
rfk: nit: suggest naming this to be super clear *which* cache is being cleared, e.g. "clearModuleCache".

rfk:
I like the direction you're taking here, but we should get @seanmonstar to have a look over it as well now that he's back.  @seanmonstar r?

tasks/test.js
philbooth: Oops, thanks.

Issue #308, opened 2015-07-20
Experiencing small number for expiredToken errors for /v1/verify


jrgm:
See this custom Kibana graph. 

https://kibana.fxa.us-west-2.prod.mozaws.net/index.html#dashboard/temp/AU6pzz-c3q0_6xopmbpt

The first error is at '2015-07-18T23:27:03.694Z'. According to my notes, I switched DNS to use the train-41 fxa-oauth stack that had a 48 hour expiry at '2015-07-16T23:18Z', so as soon as 2-day tokens started expiring.

I subsequently re-deployed fxa-oauth with a 2-week expiry at '2015-07-18T00:35Z', so 2-day tokens stopped being issued then (modulo DNS laggards). 

The error rate is about 0.5% overall, and a bit higher more recently. It's unclear if this is beginning to diminish, but maybe. 

There's not much more detail in the logs than in that graph. I could capture a bit more information on the wire if we need that. I could also update the records that have expired in the database, but that's a sucky query and may cause a little service disruption. I could also deploy a new stack to set new token expirations further in the future than 2 weeks, while we investigate. 

Anyways, thoughts?

rfk:
I suspect this will be either desktop or android clients who are fetching profile images using implicit-grant tokens, and caching those tokens until they expire.  We should add more info to the logs for debugging purposes, e.g. the client_id that owns the token.

rfk:
Next step here is to wait for #309 to hit production in train-42, and verify the identities of the clients sending expired tokens.

jrgm:
I grabbed a little network traffic to get some information on the expired tokens. Of course, duh. They are expired tokens, so they don't exist anymore. 

The only bit of information I got was from the http response, which noted they had expired sometime Sunday and only that day, which is the expected answer (2-day tokens stopped being issued on Friday).

rfk:
Oh, do we delete them from the database when expired?  I thought we hadn't implemented the pruning part yet.

seanmonstar:
They aren't be deleted.

jrgm:
New OAuth stack deployed and in production at ~15:24 PDT https://bugzilla.mozilla.org/show_bug.cgi?id=1189958

rfk:
Closing this in favour of https://github.com/mozilla/fxa-oauth-server/issues/365, which will concretely track our attempt to bring token expiry back to life.

PR #309, opened 2015-07-20
chore(logging): Log additional details for debugging expired tokens

rfk:
Adds logging to help diagnose who is using expired tokens in #308.

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -7,6 +7,9 @@ const auth = require('./auth');
 const db = require('./db');
 const encrypt = require('./encrypt');
 const Scope = require('./scope');
+const logger  = require('./logging')('token');
+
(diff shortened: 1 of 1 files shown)


rfk:
@jrgm here's a suggestion for getting more details on #308.  It adds logging not only if someone tries to use an expired token, but if someone uses a token that's more than 24 hours old.  This should give us info on reliers that might be burnt by reducing the access_token expiry time in the future.

I haven't added any tests because we're not set up for it; this would be another fine application of the mocking stuff @philbooth added in #305.

Thoughts?

jrgm:
Yeah, haven't run it, but LGTM. Probably should do this in train-42. (I marked some other pending PR to be train-43). Thanks.

rfk:
@seanmonstar r?

jrgm:
r+

Issue #310, opened 2015-07-22
Check that account deletions are being properly processed in production


rfk:
We definitely have some code for this, but I'm not sure if we got it all the way through to production.  @seanmonstar please check it out.

rfk:
@seanmonstar I recall we landed a patch to enforce some of this, is there any work remaining to do on this bug?

seanmonstar:
Technically, I don't think the config is set up yet, but I don't know what the values would be. However, the merged patched will not allow a deployment (to prod or stage) without that config being filled in.

PR #312, opened 2015-07-23
fix(events): require events to be configured in production

seanmonstar:
While abrupt, forgetting to delete user information is a big deal, so
this will prevent the server from running as long as the event queue is
not setup.

BREAKING CHANGE: Server will fail to start up if `config.events` is not
  set with values when in production.

See #310 

cc @ckolos: I noticed the puppet-config is missing these values, so they will need to be updated.

### DIFF
--- a/lib/env.js
+++ b/lib/env.js
@@ -1,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const config = require('./config');
+
+exports.isProdLike = function isProdLike() {
+  var env = config.get('env');
+  return env === 'prod' || env === 'stage';
+};
--- a/lib/events.js
+++ b/lib/events.js
@@ -4,6 +4,7 @@
 
 const config = require('./config').root();
 const db = require('./db');
+const env = require('./env');
 const logger = require('./logging')('events');
 const Sink = require('fxa-notifier-aws').Sink;
 const HEX_STRING = require('./validators').HEX_STRING;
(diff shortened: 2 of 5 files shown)


rfk:
r+; but we'll need to coordinate with @jrgm for when to merge and roll this out

lib/events.js
```diff
@@ -12,12 +12,19 @@ var fxaEvents;
 
 if (!config.events.region || !config.events.queueUrl) {
   fxaEvents = {
-    start: function () { logger.warn('accountEvent.unconfigured'); }
+    start: function start() {
+      if (config.env === 'prod') {
```
jrgm: Could you change this to 
```
if (config.env === 'prod' || config.env === 'stage') {
```
Thanks.

rfk: We have `  var isProd = config.env === 'prod';` in other others of the code, should we update those to include stage as well while we think of it?

jrgm: Yeah, I think so. You want stage to behave as much as possible like production.

seanmonstar:
Ah, I'm still not used to the change that @jrgm is ops now. I'll stop poking ckolos.

Anywho, I've updated all instances `env === 'prod'` to a `env.isProdLike()`.

lib/env.js
```diff
@@ -0,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const config = require('./config');
+
+exports.isProdLike = function isProdLike() {
+  var env = config.get('env');
+  return env === 'prod' || env === 'stage';
```
jrgm: Oh right. fxa-dev will need an update to set a config of events since latest.dev.lcip.org runs as 'env === "stage"', and I guess an SQS (or maybe we can just dummy that on those boxes).

seanmonstar: Optionally, another route is to move this requirement away from `isProdLike`, and instead add a `config.events.required` value, default to true, and if `env` is `dev` or `test`, switch to false. Then, latest could also just set this requirement to false... :fried_shrimp:

seanmonstar:
I was assigned, but what's the action? Shall I merge, or change to a `config.events.required` approach?

rfk:
I think the current approach is fine, merging

PR #316, opened 2015-07-27
fix(authorization): allow empty scope with implicit grant

### LINKED ISSUE #315: missing scope parameter causes a CRITICAL error to /v1/authorization
Found while tracking down problems related to #314.

If I call POST /authorization w/o a `scope`, a critical error is logged:

> oauth-server PORT 9010-6 (out): fxa-oauth-server.server.hapi.CRITICAL: error.uncaught {"tags":["hapi","internal","implementation","error"],"error":{"isBoom":true,"output":{"statusCode":500,"payload":{"statusCode":500,"error":"Internal Server Error","message":"An internal server error occurred"},"headers":{}},"isDeveloperError":true}}
> oauth-server PORT 9010-6 (out): fxa-oauth-server.summary.ERROR: summary {"code":500,"errno":999,"method":"post","path":"/v1/authorization","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:42.0) Gecko/20100101 Firefox/42.0","t":8,"client_id":"98e6508e88680e1a","auth":null,"payload":["client_id","assertion","response_type","ttl","access_type","redirect_uri"],"stack":"Error: scope is not allowed to be empty\n    at Object.exports.create (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/node_modules/boom/lib/index.js:21:17)\n    at Object.exports.internal (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/node_modules/boom/lib/index.js:246:92)\n    at Object.exports.badImplementation (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/node_modules/boom/lib/index.js:282:23)\n    at postValidate (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/lib/validation.js:150:26)\n    at internals.Any._validateWithOptions (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/joi/lib/any.js:429:16)\n    at root.validate (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/joi/lib/index.js:102:23)\n    at exports.response (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/lib/validation.js:159:16)\n    at /Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/lib/request.js:304:13\n    at iterate (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/node_modules/items/lib/index.js:35:13)\n    at done (/Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/node_modules/items/lib/index.js:27:25)\n    at /Users/stomlinson/development/fxa-local-dev/fxa-oauth-server/node_modules/hapi/node_modules/hoek/lib/index.js:756:22\n    at process._tickDomainCallback (node.js:463:13)"}

The request body:
```json
{
  "client_id": "98e6508e88680e1a",
  "assertion":...,
  "response_type":"token"
}
```

The response body:
```json
{
  "code": 500,
  "errno": 999,
  "error": "Internal Server Error",
  "message": "An internal server error occurred",
  "info": "https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#errors"
}
```

expected:
No nasty stack trace with a message about an uncaught error at a minimum, possibly a missing parameter error returned to the client similar to the auth server.

seanmonstar:
Closes #315

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -55,13 +55,17 @@ exports.create = function createServer() {
   // hapi internal logging: server and request
   server.on('log', function onServerLog(ev, tags) {
     if (tags.error && tags.implementation) {
-      hapiLogger.critical('error.uncaught', { tags: ev.tags, error: ev.data });
+      hapiLogger.critical('error.uncaught.server', ev.data);
     }
   });
 
   server.on('request', function onRequestLog(req, ev, tags) {
     if (tags.error && tags.implementation) {
-      hapiLogger.critical('error.uncaught', { tags: ev.tags, error: ev.data });
+      if (ev.data.stack.indexOf('hapi/lib/validation.js') !== -1) {
+        hapiLogger.error('error.payload.validation', ev.data);
+      } else {
+        hapiLogger.critical('error.uncaught.request', ev.data);
+      }
     }
   });
 
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -139,7 +139,7 @@ module.exports = {
       redirect: Joi.string(),
       access_token: validators.token,
       token_type: Joi.string().valid('bearer'),
-      scope: Joi.string(),
+      scope: Joi.string().allow(''),
       auth_at: Joi.number(),
       expires_in: Joi.number()
     }).without('redirect', [
--- a/tasks/release.js
+++ b/tasks/release.js
@@ -22,8 +22,6 @@ module.exports = function (grunt) {
     dest: 'CHANGELOG.md'
   });
 
-  grunt.renameTask('release', 'version');
-
   grunt.registerTask('version', [
     'bump-only:minor',
     'changelog',
--- a/test/api.js
+++ b/test/api.js
@@ -556,6 +556,21 @@ describe('/v1', function() {
             assert.equal(res.statusCode, 200);
           });
         });
+
+        it('does not require scope argument', function() {
+          mockAssertion().reply(200, VERIFY_GOOD);
+          return Server.api.post({
+            url: '/authorization',
+            payload: authParams({
+              client_id: client2.id,
+              scope: undefined,
+              response_type: 'token'
+            })
+          }).then(function(res) {
+            assert.equal(res.statusCode, 200);
+          });
+        });
+
         it('requires a client with proper permission', function() {
           mockAssertion().reply(200, VERIFY_GOOD);
           return Server.api.post({


rfk:
What does it mean for someone to have an oauth token with an empty scope?  What capabilities or permissions does it grant them?  I wonder if this always represents a client error of some kind.

shane-tomlinson:
> What does it mean for someone to have an oauth token with an empty scope? What capabilities or permissions does it grant them? I wonder if this always represents a client error of some kind.

This was my question too. The non-implicit flow allows a non-existent scope, which seems to do exactly nothing. See https://github.com/mozilla/fxa-content-server/issues/2832#issuecomment-125180301

seanmonstar:
A token with no scope can't really do anything, such that a scope of `''` and a scope of `'carebears'` would be equally useless, and equally allowed.

rfk:
> The non-implicit flow allows a non-existent scope

I agree we should be consistent then, but IMHO we should disallow it in both places.  No good can come from a client requesting a token with no scopes.  Would it break any clients if we started erroring out on empty tokens in the explicit flow?

seanmonstar:
Just thinking out loud here: a token with no scopes might be useful to
prove that a user has an account, and can authenticate it, but otherwise
the site doesn't care who exactly it is?

rfk:
I remain unconvinced, but will defer to your judgement @seanmonstar.  Certainly nothing *bad* will happen if we allow the client to request  scopeless token, most likely they'll just get an error slightly later down the chain when they appempt to use it for something.

For consistency, we should also change refresh-token handling to allow narrowing the scope of an existing token down to nothing.

seanmonstar:
I wasn't that convinced, myself. The spec says the decision is up to us:
https://tools.ietf.org/html/rfc6749#section-3.3

ckarlof:
As a untrusted relier, if you want to reduce friction in the login experience, you could request no scopes,  and still get the stable uid in https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1verify. It's unclear to me whether we skip the permission screen when the relier requests no scopes, but this still presents an opportunity to explain WTF is going on.

rfk:
> you could request no scopes, and still get the stable uid

We have a "profile:uid" scope that in theory mediates access to this, although I don't recall how heavy-handedly we enfore that.

seanmonstar:
It's technically possible to get the uid without the scope, if they pass the token to oauth server's `/verify` endpoint. This is because any service provider would need to know what uid it is dealing with.

lib/server/index.js
```diff
@@ -55,13 +55,17 @@ exports.create = function createServer() {
...
     if (tags.error && tags.implementation) {
-      hapiLogger.critical('error.uncaught', { tags: ev.tags, error: ev.data });
+      hapiLogger.critical('error.uncaught.server', ev.data);
     }
   });
 
   server.on('request', function onRequestLog(req, ev, tags) {
     if (tags.error && tags.implementation) {
-      hapiLogger.critical('error.uncaught', { tags: ev.tags, error: ev.data });
+      if (ev.data.stack.indexOf('hapi/lib/validation.js') !== -1) {
```
rfk: maybe worth  comment to say why we're introspecting the traceback stack here

rfk:
Alright, let's ship it then.  I think we should do this for consistency:

> we should also change refresh-token handling to allow narrowing the scope of an existing token down to nothing.

PR #318, opened 2015-07-30
fix(db): we need to enforce only a minimum patch level (not {n,n+1})

jrgm:
Hey, @seanmonstar - this is similar to the change in patch level restrictions we made for fxa-auth-db-mysql. It isn't always true that there is one db patch in each train, and train-42 has two (that amount to a no-op, but that's a different story). 

I will have to backport this to v0.41.1, where it will be slightly different due to other changes to this file. 
But first, I wanted to get this in master. 

r?

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -70,16 +70,17 @@ function checkDbPatchLevel(patcher) {
     if (err) {
       return d.reject(err);
     }
-    // We are only guaranteed to run correctly if we're at the current
-    // patch level for this version of the code (the normal state of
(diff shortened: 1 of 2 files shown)


seanmonstar:
This looks fine, but we do we need to backport to 0.41.1? That patch doesn't have any DB changes...

jrgm:
So, v0.41.0 would have needed a change too (to be safe against a restart). I haven't yet made the no-op bump in version number in production yet. But when I do, the oauth server, should it restart, would fail to restart. v0.41.n expect patch level 6, but that will move to 8 in train-42.

PR #325, opened 2015-08-17
fix(db): don't change client database at startup; footgun

jrgm:
To change secrets is a precarious dance with live servers. It would be better if this could did not muck with the database at startup. 

r? - @seanmonstar

### DIFF
--- a/lib/db/index.js
+++ b/lib/db/index.js
@@ -8,6 +8,7 @@ const P = require('../promise');
 
 const config = require('../config');
 const encrypt = require('../encrypt');
+const env = require('../env');
 const logger = require('../logging')('db');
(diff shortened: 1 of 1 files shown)


lib/db/index.js
```diff
@@ -12,6 +12,11 @@ const logger = require('../logging')('db');
 const klass = config.get('db.driver') === 'mysql' ?
   require('./mysql') : require('./memory');
 
+function isProduction() {
```
seanmonstar: use `require('../env').isProdLike()`

jrgm:
> use require('../env').isProdLike()

Thanks. Done rebased and pushed.

seanmonstar:
Looks like a spurious travis failure, since all passed but the 1. Retrying that job.

PR #326, opened 2015-08-19
chore(docs): add a note about dev envs

### LINKED ISSUE #148: Add documentation that any localhost and 127.0.0.1 redirect_uri can be used in our dev environments
Both to our API docs in the repo and in our MDN article:

https://developer.mozilla.org/en-US/Firefox_Accounts

vladikoff:
Fixes #148


rfk:
LGTM, although it feels like maybe it should be on its own line rather than part of the preceeding paragraph.

PR #333, opened 2015-09-08
chore(version): generate legacy-format output for ./config/version.json

jrgm:
To add the source repo to the output of /__version__ in production rpms I need to have this script in the tree. For train-45.


rfk:
:+1:

PR #335, opened 2015-09-11
feat(clients): add notion of Service Clients in config

### LINKED ISSUE #327: Registering Service Accounts
The first part of https://github.com/mozilla/fxa/issues/53, an ability to create client that can have a public/private keypair generated, with the public key saved in a configuration file. Since this requires explicit registration with out help, use of a database isn't quite needed.

As this client doesn't need most of the properties that a typical client does, create a new config value, `service_clients`, with the properties `client_id`, `pubkey`, `scope`.

seanmonstar:
Closes #327


seanmonstar:
@rfk r? or should i pick someone else?

lib/config.js
```diff
@@ -173,6 +173,10 @@ const conf = convict({
       default: 9011
     }
   },
+  serviceClients: {
+    doc: 'Service Clients',
```
rfk: This feels like it needs a slightly more detailed docstring :-)

config/test.json
```diff
@@ -49,5 +49,17 @@
...
+  },
+  "serviceClients": [
+    {
+      "id": "d23dbf62b82eb04e",
+      "name": "Test Service Client",
+      "scope": "profile:email",
+      "pubkey": {
+        "kty": "RSA",
+        "n":"xaQHsKpu1KSK-YEMoLzZS7Xxciy3fsGrhrrqW_JBrq3IRmeGLaqlE80zcpIVnStyp9tbet2niYTemt8ug591YWO5Y-S0EgQyFTxnGjzNOvAL6Cd2iGie9QeSehfFLNyRPdQiadYw07fw-h5gweMpVJv8nTgS-Bcorlw9JQM6Il1cUpbP0Lt-F_5qrzlaOiTEAAb4JGOusVh0n-MZfKt7w0mikauMH5KfhflwQDn4YTzRkWJzlldXr1Cs0ZkYzOwS4Hcoku7vd6lqCUO0GgZvkuvCFqdVKzpa4CGboNdfIjcGVF4f1CTQaQ0ao51cwLzq1pgi5aWYhVH7lJcm6O_BQw",
+        "e":"AQAB"
```
rfk: We'll need a way to support rotating these keys, so at the very least, it should probably be a set of keys rather than a single item.  In the absence of reasons to deviate from spec, we could use a JWK set [1].

I know I harp on about it, but I think we should consider using JKUs [2] here, and fetching the public keys dynamically rather than embedding them directly in the config.  This seemed to work nicely for preVerifyTokens in the auth-server, and provided a nice decoupling between systems.

Pros:
* decouples services from our config here, we just store the JKU value that we expect to see from them, and look up the actual key material dynamically
* services can rotate their keys without us having to change our config at all

Cons:
* Will probably make testing and local development harder, as you have to be able to fetch the keys
* Reduces the security of the system to that of TLS (although we can mitigate with cert pinning, and by ensuring the services don't communicate over the open internet)

@seanmonstar I'd be interested in your thoughts either way.

[1] http://tools.ietf.org/html/rfc7517#section-5
[2] http://tools.ietf.org/html/rfc7515#section-4.1.2

seanmonstar: A JWK set makes perfect sense here. Looking at JKUs, and their use in the auth-server, I don't see pros. It looks like the auth-server is fetching the JWK set over HTTPS... that is a route served by the exact same server? How, then, is this decoupled from the config?

dannycoates: I might not be following this closely enough to dive in here, but regarding the auth-server an JKUs, here's how it works:

@seanmonstar is the route you're talking about `/.well-known/public-keys`? If so, that's basically a "demo" route to show what a jku route should return and not used by preVerifyTokens (PVT) at all. 

What PVTs do use is a config list of `trustedJKUs` which get fetched as needed when specified in a PVT instead of having the jwks in config. In practice we only had one trusted JKU, marketplace, but it allowed for them to manage their keys on their servers without us being involved once we knew their JKU endpoint. If they ever changed them, we'd notice that we're missing a key with that id from the PVT and refetch. I hope that makes sense.

The key benefit is that we only needed to maintain the trusted JKU list, and not the actual keys.

seanmonstar: Oh, that makes sense. Look up the JWK set from basket's server (or whoever else is a service client). I thought you were looking it up on the same server.

seanmonstar:
@rfk ok, so we can use JKUs. I've updated the config here to show that. Usage of that JKU value in the test config will be in the other PR, with the request mocked out using `nock`.

rfk:
LGTM

PR #336, opened 2015-09-11
Service Client Tokens

### LINKED ISSUE #328: Accept JWT from Service Clients and return authorization token
Following [Google's lead](https://developers.google.com/identity/protocols/OAuth2ServiceAccount), Service Clients will generate JWTs to request data for a specific user, signing it with the private key that was gotten during registration.

Algorithms:

- RSA SHA-256 (wide support)
- [ED25519](http://ed25519.cr.yp.to/) (**wishlist**. smaller, faster, and more secure than RSA. less library support)

Example JavaScript of making the JWT:

```js
var now = Math.floor(Date.now() / 1000);
var header = { alg: 'RS256', typ: 'JWT' };
var claims = {
  iss: clientId,
  scope: 'profile:email',
  aud: 'https://oauth.accounts.firefox.com/v1/authorization',
  iat: now,
  exp: now + (60 * 5),
  sub: userId + '@accounts.firefox.com'
};

var token = base64(JSON.stringify(header)) + '.' + base64(JSON.stringify(claims));
var sig = rsa256(Buffer(token, 'base64'), privateKey);
var jwt = token + '.' + base64(sig);
```

The `jwt` calculated above would be used as an argument when requesting an authorization token.

```
POST /v1/authorization?grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion=jwt
```

No additional parameters are required (and are ignored (illegal?)), as they are contained within the signed assertion claim set. The authorization_token will expire, just as when a typical client uses a refresh token.

### LINKED ISSUE #329: Document Service Clients
- [x] Documentation about how to register a new Service Client
- [x] Update documentation on `POST /v1/authorization`, explaining how to use a JWT

seanmonstar:
This currently works, posting an assertion to `/v1/token` with the right properties in the token will give you an access token.

**One potential blocker** that has shown up implementing this:

- Since the token is requested without an fxa-assertion, we cannot store the user's email address with the token, and therefore cannot fulfill requests to the profile server for `/v1/email`.

Closes #328 
Closes #329


GitCop:

 * Commit: 5889eb5a9b73bd7da1dade24c7bb02913b6ef73a

---

rfk:
> Since the token is requested without an fxa-assertion, we cannot store the user's email address
> with the token, and therefore cannot fulfill requests to the profile server for /v1/email

That's going to get a bit messy, but TBH I'm glad this has bitten us so squarely because it needs to be refactored away sooner or later anyway.  The right thing is for us to directly ask the auth-server for the email associated with the account, passing the OAuth bearer token along to authorize the request.

rfk:
> Since these clients are only in config, the foreign key restriction on access_tokens.clientId will prevent
> inserting these tokens into the MySQL database.

Based on discussion today, ISTM that we either:

1.  Create stub entries for the service clients in the `clients` table, but keep their special JWK stuff separate so that there's no risk of accidentally allowing other clients to act as service clients.
2.  Just drop the foreign key constraint and be done with it.

I feel like the constraint has some value so I'd lean slghtly towards (1), but don't feel strongly.  @seanmonstar thoughts?

GitCop:

 * Commit: d9357512901c733a2d15b72f7a317f01f662a167

---

seanmonstar:
Woo! tests passed! As in... the missing email field won't blow up the mysql db. It still won't be able to respond with an email until https://github.com/mozilla/fxa-auth-server/pull/915 is merged.

lib/db/mysql/schema.sql
```diff
@@ -41,7 +41,7 @@ CREATE TABLE IF NOT EXISTS tokens (
   FOREIGN KEY (clientId) REFERENCES clients(id) ON DELETE CASCADE,
   userId BINARY(16) NOT NULL,
   INDEX tokens_user_id(userId),
-  email VARCHAR(256) NOT NULL,
+  email VARCHAR(256),
```
rfk: I don't know about this intermediate state where some tokens can't fetch the email address...ISTM maybe we should go through with something like https://github.com/mozilla/fxa-auth-server/issues/1053 and then we can just drop this column entirely.

In fact, I wonder if we can deprecate having the email returned implicitly by /verify, and it's a bit of an abstraction violation anyway...

lib/routes/token.js
```diff
@@ -16,9 +17,20 @@ const P = require('../promise');
...
 const validators = require('../validators');
 
+const HEX_STRING = validators.HEX_STRING;
+
 const MAX_TTL_S = config.get('expiration.accessToken') / 1000;
 const GRANT_AUTHORIZATION_CODE = 'authorization_code';
 const GRANT_REFRESH_TOKEN = 'refresh_token';
+const GRANT_JWT = 'urn:ietf:params:oauth:grant-type:jwt-bearer';
+
+const JWT_AUD = config.get('publicUrl') + '/v1/token';
```
rfk: Is it usual to have audiences with a path component?  I don't recall seeing this pattern used elsewhere.  I guess it can't hurt though...

lib/routes/token.js
```diff
@@ -89,6 +101,60 @@ function confirmRefreshToken(params) {
...
+      logger.debug('jwt.invalid.aud', payload.aud);
+      throw AppError.invalidAssertion();
+    }
+
+    if (!Scope(client.scope).has(payload.scope)) {
+      logger.debug('jwt.invalid.scopes', {
+        allowed: client.scope,
+        requested: payload.scope
+      });
+      throw AppError.invalidScopes();
```
rfk: Pass the scopes to the constructor, for client-visible error message.

lib/routes/token.js
```diff
@@ -89,6 +101,60 @@ function confirmRefreshToken(params) {
...
+    }
+
+    var now = Date.now() / 1000;
+    if ((payload.iat || Infinity) > now) {
+      logger.debug('jwt.invalid.iat', { now: now, iat: payload.iat });
+      throw AppError.invalidAssertion();
+    }
+    if ((payload.exp || -Infinity) < now) {
```
rfk: IMHO we should flatly refuse to accept JWTs without an explicit expiry timestamp.

lib/routes/token.js
```diff
@@ -89,6 +101,60 @@ function confirmRefreshToken(params) {
...
+        requested: payload.scope
+      });
+      throw AppError.invalidScopes();
+    }
+
+    var now = Date.now() / 1000;
+    if ((payload.iat || Infinity) > now) {
+      logger.debug('jwt.invalid.iat', { now: now, iat: payload.iat });
+      throw AppError.invalidAssertion();
+    }
```
rfk: Clock skew has caused similar checks to fail surprisingly often.  Since this is for server-to-server communication, we can assume reasonably well-synchronised clocks, but I think we should still add some small window for clock skew between the two systems.

lib/routes/token.js
```diff
@@ -89,6 +101,60 @@ function confirmRefreshToken(params) {
...
+  var assertion = params.assertion;
+  logger.debug('jwt.confirm', assertion);
+
+  return JWTOOL.verify(assertion).catch(function(err) {
+    logger.info('jwt.invalid.verify', err.message);
+    throw AppError.invalidAssertion();
+  }).then(function(payload) {
+    logger.verbose('jwt.payload', payload);
+
+    var client = SERVICE_CLIENTS[payload.iss];
```
rfk: IIUC, the `iss` field here is the client's opaque hex id.  I think it would be quite neat if the client didn't need to know this value, and could just put their domain-name in directly like `iss: basket.mozilla.org`, in a similar manner to what we do for identity assertions.  One less piece of coupling between the two systems.

That might complicate the data model though.

rfk:
@seanmonstar by the way I found [1] while googling for the `urn:` string used here, is the API here based on that document, or just from the google API?

[1] http://self-issued.info/docs/draft-ietf-oauth-jwt-bearer.html

lib/routes/token.js
rfk: >  I don't recall seeing this pattern used elsewhere

It's used in both [1] and [2] so I guess I just haven't been paying attention.  :+1: 

[1] https://developers.google.com/identity/protocols/OAuth2ServiceAccount
[2] http://self-issued.info/docs/draft-ietf-oauth-jwt-bearer.html

lib/routes/token.js
```diff
@@ -172,18 +256,28 @@ module.exports = {
...
+        return confirmJwt(params);
+      } else {
+        // else our Joi validation failed us?
+        logger.critical('joi.grant_type', params.grant_type);
+        throw Error('unreachable');
+      }
+    })
+    .then(function(vals) {
+      vals.ttl = params.ttl;
+      return vals;
```
rfk: This file's getting pretty long and growing a lot of concerns; I think it would benefit from a few high-level comments to assist new readers in navigating the file.

lib/routes/token.js
```diff
@@ -89,6 +101,60 @@ function confirmRefreshToken(params) {
...
+    if (!client) {
+      logger.debug('jwt.client.unknown', payload.iss);
+      throw AppError.unknownClient(payload.iss);
+    }
+
+    var uid = payload.sub;
+    if (!uid || !(uid.length === 32 && HEX_STRING.test(uid))) {
+      logger.debug('jwt.invalid.sub', uid);
+      throw AppError.invalidAssertion();
+    }
```
rfk: In #328 you proposed `"id@accounts.firefox.com"` for the `sub` field, but this just uses the raw uid.  Any comments on which is superior and why?

(FWIW I'm generally in favour of using namespaced identifeirs, but there's no real ambiguity here so I think the plain id is fine as well.  The only thing we'd do if you gave us an unexpected domain in the identifier would be return an error)

rfk:
@seanmonstar this is looking good.  I think we should also have #329 ready before merging though, either as part of this PR or separately, since it will make it easier to think through the shape of the feature as a whole.

lib/routes/token.js
rfk: Also, it's a bit strange that we don't get this expiration-time checking for free from `fxa-jwtool`.  I wonder if there's an API we could add to that library to simplify some of the work here, e.g. a `validate()` method that we can pass in a timestamp and aud and have it enforce these checks for us.  @dannycoates thoughts?

lib/db/mysql/schema.sql
seanmonstar: Yea, that'd be ideal. However, I don't know my way around the auth server much, and to completely drop this column, the auth server would need that PR merged, or else **ALL** requests for `email` would no longer be serviceable.

I didn't want to block landing this based on that, but if it's possible to get that landed asap, then we could just drop all the `email` columns in the oauth db.

lib/routes/token.js
seanmonstar: This does that, without introducing another branch (which jscs likes to complain about complexity). If there's no `payload.exp`, then `-Infinity` is used, and `-Infinity < now` will always be true, such that this would be considered expired.

lib/routes/token.js
seanmonstar: I've never understood the clock skew problem that well. I know what it is, it just seems that if there is an issue, just push expiration out another couple minutes or something?

lib/routes/token.js
seanmonstar: That's probably doable.

lib/routes/token.js
seanmonstar: btw I never want to open this file again

lib/routes/token.js
seanmonstar: I don't think 1 is really superior to the other. I think I proposed that because I was looking at the existing `browserid` module used with fxa's browserid assertions, and then proceeded to forget about when writing the code.

lib/routes/token.js
rfk: Ah yes, right you are

lib/routes/token.js
seanmonstar: Ok, here's how I'm doing this: I don't even check for a `iss` claim anymore.

The verification won't work if the assertion doesn't include a `jku` in the header that we don't already trust. The current design is that each client has only 1 jku, so I can look up the client based on the jku in the assertion. The only caveat: this requires that every client have a unique jku. No 2 clients can share one. This seems reasonable to me...

seanmonstar:
@rfk updated from comments. Added commit that includes documentation on Service Clients.

docs/service-clients.md
```diff
@@ -0,0 +1,92 @@
...
+and then send it to our [/v1/token][] endpoint.
+
+Here's an example of creating a JWT in JavaScript:
+
+```js
+var now = Math.floor(Date.now() / 1000); // in seconds
+var header = {
+  alg: 'RS256',
+  typ: 'JWT',
+  jku: 'https://basket.accounts.firefox.com/.well-known/jku',
```
rfk: nit: may as well use the real basket domain name of `basket.mozilla.org`

lib/db/index.js
```diff
@@ -108,6 +109,32 @@ function preClients() {
...
+      return exports.getClient(client.id).then(function(existing) {
+        if (existing) {
+          logger.verbose('seviceClients.existing', client);
+          return;
+        }
+
+        return exports.registerClient({
+          id: client.id,
+          name: client.name,
+          hashedSecret: encrypt.hash(unique.secret()),
```
rfk: In other places we've used "0000000000000000000000000000000000000000000000000000000000000000" as the hashedSecret for clients that aren't supposed to use a secret.  Is that a pattern we should repeat for these service clients for consistency?

seanmonstar: No, that pattern was actually a poor idea, and I clarified in bugs that
those should NOT be the hashes for those clients (cause something can
indeed hash to 0), but that the clients should have random hashes that were
thrown away afterwards.

However, this inserts into the db at startup, which @jrgm may object too.

docs/service-clients.md
```diff
@@ -0,0 +1,92 @@
...
+  kid: 'k1'
+};
+var claims = {
+  scope: 'profile:email',
+  aud: 'https://oauth.accounts.firefox.com/v1/token',
+  iat: now,
+  exp: now + (60 * 5),
+  sub: userId + '@accounts.firefox.com'
```
rfk: nit: I think this will have to be "@api.accounts.firefox.com" in production, which is pretty ugly, but it's what we have...

lib/routes/token.js
```diff
@@ -16,9 +19,146 @@ const P = require('../promise');
...
+//      confirm the client credentials in `confirmClient()`.
+//      - If grant type is authorization code, proceed to `confirmCode()`.
+//      - If grant type is refresh token, proceed to `confirmRefreshToken()`.
+//    - If grant type is a JWT, all information is in the JWT. So jump
+//      straight to `confirmJwt()`.
+// 2. Generate tokens.
+//   - An options object is passed to `generateTokens()`.
+//     - An access_token is generated.
+//     - If req.payload.access_type = 'offline', a refresh_token is also
+//       generated.
```
rfk: It sounds like this will allow all grant types to generate refresh tokens.  I wonder if we should restrict it to only `authorization_code` clients.  Service clients have no legitimate use for a refresh token, since they can generate access tokens on demand, and having them able to generate refresh tokens could make it harder to e.g. thoroughly revoke service client access.

lib/routes/token.js
rfk: It can also be a problem with `iat`.  If the service-client's clock is slightly ahead of the oauth-server's clock, it can generate an assertion that appears to the oauth-server to be from the future, and hence will be rejected.  In other situations we've worked around this by checking like `if (iat || Infinity > now - 5)` just to add a bit of leeway.

lib/routes/token.js
```diff
@@ -89,6 +229,75 @@ function confirmRefreshToken(params) {
...
+  if (uid.length !== 32 ||
+    !HEX_STRING.test(uid) ||
+    ns !== allowedNs ||
+    parts.length !== 2) {
+    logger.debug('jwt.invalid.sub', sub);
+    throw AppError.invalidAssertion();
+  }
+
+  return uid;
+}
```
rfk: After sleeping on it, I lean towards allowing raw FxA uids here rather than requiring them to be namespaced.  It's the form we've been encouraging clients and reliers to use in other places.

rfk:
> However, this inserts into the db at startup, which @jrgm may object too.

I'm going to kick this over to @jrgm for comment on that one.  It seems to me that a reasonable compromise would be to auto-create any clients that do not exist in the database, but not attempt to modify any existing client records in the db. (instead logging a warning for manual review).

lib/routes/token.js
seanmonstar: Hm, woops. This line is inaccurate. You cannot pass `access_type` to this route. I'll adjust this line to point out that a refresh token is generated if the authorization_code was created with offline access.

lib/routes/token.js
seanmonstar: It has been? Where's an example? I settled on using namespaces, cause they don't hurt, and it's backwards compatible to eventually drop the namespace, but potentially confusing if we need to add one down the road.

seanmonstar:
> but not attempt to modify any existing client records in the db. (instead logging a warning for manual review).

Current code looks up via the client id in the config, and if found, continues the loop (not inserting). It will never edit an exist record. This will happen at **every** process startup, so logging a warning would happen every time after the first, and weaken the strength of warnings. Also, there shouldn't be anything to manually edit, since the only reason for it to be in the database is for the foreign key restriction. Any changes we might need to make would be in the config (changing the `scope` or `jku`).

rfk:
> This will happen at every process startup, so logging a warning would happen 
> every time after the first, and weaken the strength of warnings

I meant log a warning iff the config and the db differ, but yeah, if there's nothing really in the db to edit anyway then it's a bit of a waste of time.

lib/routes/token.js
rfk: It's what we tell reliers to store in their db as the account primary key, for example.  Basket and pocket both store it as "<uid>" rather than "<uid>@api.accounts.firefox.com".

seanmonstar: Ok, I'll rip the namespaces back out :)

rfk: The other alternative it so accept both forms, but we should probably just keep it simple :-)

seanmonstar:
@rfk updated to use plain userids, instead of a namespace.

rfk:
We chatted with @jrgm briefly today and got tentative support for the "insert but don't modify" approach to handling these clients on startup; assigning back to myself for final review

docs/api.md
```diff
@@ -368,20 +368,26 @@ particular user.
...
 
   The default and maximum value is 2 weeks.
-- `grant_type`: Either the string `authorization_code` or `refresh_token`.
+- `grant_type`: Either `authorization_code`, `refresh_token`, or `urn:ietf:params:oauth:grant-type:jwt-bearer`.
   - If `authorization_code`:
+    - `client_id`: The id returned from client registration.
+    - `client_secret`: The secret returned from client registration.
     - `code`: A string that was received from the [authorization][] endpoint.
   - If `refresh_token`:
+     - `client_id`: The id returned from client registration.
```
rfk: nit: extra indentation here

lib/routes/token.js
rfk: > if (iat || Infinity > now - 5)

Re-reading this, it should be `+ 5` for "was issued more than five seconds in the future".  I still think it's worth adding some small allowance for skew here, we have seen this cause problems in other systems in the past.

rfk:
r+ with a couple of final nits; we'll need a follow-up bug for using https://github.com/mozilla/fxa-auth-server/pull/1070 once it's landed, but assuming the `MODIFY COLUMN` is not a painful migration, we can land this chunk of work in the meantime.

seanmonstar:
The failure is because a test in db/mysql is calling process.exit(1), and that seems to be intercepted correctly on node 0.10, but not 0.12?

PR #338, opened 2015-09-24
chore(tests): remove weird mocking magic

philbooth:
I meant to do this ages ago but forgot about it until it got mentioned in IRC the other day.

When I added the mocking stuff to this repo, I'd never used `proxyquire` before and assumed that it worked the same way as `mockery`. Of course, it doesn't.

So all the register/deregister module-cache hoop-jumping was completely unnecessary. Instead I should just have used the object returned by `proxyquire` directly.

The one thing that doesn't work using this approach is mocking the logger. So to stop those tests from failing, I've added an option for injecting the mock logger to `MysqlStore.connect`.

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -11,7 +11,6 @@ const MysqlPatcher = require('mysql-patcher');
 
 const config = require('../../config');
 const encrypt = require('../../encrypt');
-const logger = require('../../logging')('db.mysql');
 const P = require('../../promise');
 const Scope = require('../../scope');
 const unique = require('../../unique');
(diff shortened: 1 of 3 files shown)


lib/db/mysql/index.js
```diff
@@ -87,6 +88,9 @@ function checkDbPatchLevel(patcher) {
 }
 
 MysqlStore.connect = function mysqlConnect(options) {
+  if (options.logger) {
```
vladikoff: do we need to check if `options` is defined here first?

vladikoff:
Looks okay to me, @seanmonstar assigned to you to check if you are okay with mocking `logging` that way

lib/db/mysql/index.js
seanmonstar: It'd be more robust, but we didn't check before this diff, and the `options` object is freely used in the lines below.

PR #341, opened 2015-10-12
fix(config): remove 00000... from hashedSecrets

### LINKED ISSUE #339: Remove ill-advised "00000000000000000000000000000000" hashedSecret values from config files
Per https://github.com/mozilla/fxa-oauth-server/pull/336/files#r40879520 using this magic value to denote a client with no secret was a bad idea; we should remove them from the dev config files in ./config so that nobody tries to resurrect it.

seanmonstar:
Closes #339

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -53,7 +53,7 @@
     {
       "name": "Firefox",
       "id": "5882386c6d801776",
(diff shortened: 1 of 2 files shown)


seanmonstar:
@vladikoff r?

vladikoff:
Awesome!

PR #344, opened 2015-10-16
chore(clients): remove deprecated 'whitelisted' column from clients table

### LINKED ISSUE #267: [meta] drop the whitelisted column in clients table and associated code and config
this will affect fxa-dev, stage and prod dbs, fxa-oauth-console, others.

rfk:
Fixes #267, which we had to revert the first time around.  Patch number 11 is already taken by #336, so I used 12 here and will have to rebase this atop that one once it merges.  The schema document is already sans `whitelisted` because that wasn't reverted when we rolled this back the first time around.

@seanmonstar r?


seanmonstar:
r+

PR #350, opened 2015-10-21
fix(docs): note that codes are single use

### LINKED ISSUE #214: API docs should mention that codes are single-use

vladikoff:
Fixes #214


vladikoff:
@rfk r?

PR #351, opened 2015-10-28
Backout the "service tokens" feature

rfk:
With apologies to @seanmonstar, this backs out the "service tokens" feature from #336.

It turns out that the db migration to drop the "NOT NULL" constraint from "tokens.email" cannot be applied via an online "ALTER TABLE", so applying it in production would lead to significant downtime.

In discussion with @jrgm we decided that rather than burning cycles on figuring out a way to ship it, we should just back out the feature and wait for https://github.com/mozilla/fxa-auth-server/issues/1053 to land.  At that point we can drop the "email" column entirely rather than trying to modify it in-place, an operation that *can* be successfully applied in production without an outage.

There are three commits here.  The first two revert the two commits from #336.  The third adds a new db migration to patch level 13, reversing the change introduced at patch level 11 by #366.  This will ensure our various dev environments wind up with their db in the same state as production.


GitCop:

 * Commit: 0abbccdc60ae34714ed199f641bcfe6d02678ed5

 * Commit: 23643298b637f916df09167db19713f4196818d5

---

rfk:
@GitCop fine, fine, here's the commits with `revert()` blah blah in the proper form...

rfk:
@seanmonstar r? and a patch-release tag please

rfk:
also @jrgm I'd also appreciate an official r? on this from your POV

jrgm:
r+ on the backout. 

Awfully sorry @seanmonstar. 

In the end, after more work with this, it may have been feasible to make this change; but if this column is going to be discarded later it does seem better to not modify it this week, and then have to do the maintenance all over again in another while (note: feasible means acceptable disruption, not no disruption).

seanmonstar:
(╯°□°）╯︵ ┻━┻

PR #356, opened 2015-11-09
fix(travis): remove broken validate shrinkwrap

jrgm:
I tried to update to grunt-nsp but that spewed errors about extraneous modules under node_modules/grunt-nsp. So just removing this broken test to fix travis. r? @seanmonstar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_script:
 
 script:
   - npm run outdated
-  - grunt validate-shrinkwrap test
+  - grunt test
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,7 +32,6 @@ before_script:
 
 script:
   - npm run outdated
-  - grunt validate-shrinkwrap test
```
vladikoff: you just need to remove `validate-shrinkwrap`, not `test` :)

PR #357, opened 2015-11-10
chore(build): switch to grunt-nsp

### LINKED ISSUE #354: Replace grunt-nsp-shrinkwrap with grunt-nsp
I think we can use something like this:

```js
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */

module.exports = function (grunt) {
  'use strict';

  grunt.config('nsp', {
    output: 'summary',
    shrinkwrap: grunt.file.readJSON('npm-shrinkwrap.json'),
  });
};
```

rfk:
Fixes #354.

As a purely cosmetic change that I don't feel strongly about, this also renames the "tasks" directory to "grunttasks" for consistency with other repos in our cluster.

@seanmonstar r?


package.json
```diff
@@ -49,7 +49,7 @@
     "grunt-jscs": "^1.5.0",
     "grunt-mocha-test": "^0.12.2",
     "grunt-nodemon": "^0.4.0",
-    "grunt-nsp-shrinkwrap": "^0.0.3",
+    "grunt-nsp": "2.1.2",
```
vladikoff: should this be ^?

Issue #352, opened 2015-10-28
Remove ability to fetch email address from the oauth-server API


rfk:
When https://github.com/mozilla/fxa-auth-server/issues/1053 lands, tokens with "profile:email" scope can be used to fetch the user's email address directly from the auth-server.  We should do so rather than keeping a copy of it in our own db.  This will allow us to drop the "email" column from the "tokens" table, unblocking the final landing of service tokens.

seanmonstar:
Is anyone getting the user's email from the `/v1/verify` request besides the profile server? If not, that's an easy deletion. If so, they shouldn't be! Shame on them!

Also if so, then I imagine the verify route will need to be able to detect recursion, since when verify is called, it would have to ask the auth-server, passing the token, which would turn around and ask the oauth-server to verify the token (wanting to fetch the email again)...

rfk:
IIRC the basket proxy does, but that's an easy fix on our side.  Sadly, it is part of our documented API, so others could legitimately be looking for it and we'll have to move carefully if we want to remove it.

seanmonstar:
Hm, I don't see how to stop the recursion without the auth-server sending an additional parameter when it wants to verify a token...

rfk:
Agree.  We should *try* to remove it, since I really don't think anyone should be using this endpoint except for our own services.

seanmonstar:
@vladikoff I tried just chopping the email out, and found out that developers use the email address everywhere... Any change would require the console to update as well.

It looks like the developer routes require oauth authorization, and the developer's email is used from the verified token. What if instead we just keyed everything regarding developers with their FxA uid?

vladikoff:
> What if instead we just keyed everything regarding developers with their FxA uid?

@seanmonstar we need to validate that the user is `@mozilla.com` email

seanmonstar:
@vladikoff oh right. Hmm, but that validation only needs to happen once, when logging in, right?

vladikoff:
@seanmonstar Yeap!

rfk:
Having mulled this over a little, I feel like we should try to set a good habit here of moving cautiously with changes to our public-facing APIs, even if it means eating a bit of short-term complexity/ugliness internally.  I'm pretty sure that nothing outside our own systems will break if we yank `email` from the `/v1/verify` response, but let's do better than "pretty sure".

Strawman:

[...]

PR #359, opened 2015-11-10
feat(verify): add opt out parameter to verify endpoint

### LINKED ISSUE #358: Add private query param to opt out of email from verify endpoint

seanmonstar:
Any requests that don't opt out of receiving the email will log a
warning.

Closes #358

### DIFF
--- a/lib/routes/verify.js
+++ b/lib/routes/verify.js
@@ -4,13 +4,15 @@
 
 const Joi = require('joi');
 
+const logger = require('../logging')('routes.verify');
 const token = require('../token');
 const validators = require('../validators');
 
 module.exports = {
   validate: {
     payload: {
-      token: validators.token.required()
+      token: validators.token.required(),
+      email: Joi.boolean().default(true)
     }
   },
   response: {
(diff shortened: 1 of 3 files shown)


lib/routes/verify.js
```diff
@@ -22,6 +24,17 @@ module.exports = {
     }
   },
   handler: function verify(req, reply) {
-    token.verify(req.payload.token).done(reply, reply);
+    token.verify(req.payload.token).then(function(info) {
+      if (req.payload.email) {
+        logger.warn('email.returned', {
```
rfk: Tiny naming nit: it's not obvious to me what "returned" means here.  Something like "requested" may be a little more intuitive.

rfk:
@seanmonstar good stuff!  Please also add a brief note in the API docs saying that the `email` field is being deprecated and describing what we're doing here.

I'm r+ for this to land on its own, but we could also go even further here and fix #352 as part of the same change, if they seem to mesh nicely.

seanmonstar:
I'll just put that in a separate PR, since it does a different thing, and there's already a separate issue for it.

PR #360, opened 2015-11-16
reinstate service tokens

### LINKED ISSUE #353: Re-instate service tokens
We had to backout the service tokens feature over in #351.  Once we stop trying to keep a local copy of the user's email address per #352, we can bring it back and ship it.  For reference the backed-out commits were:

* 799f0e22: docs(service-clients): Document Service Clients, JKUs, and JWTs
* 0a0e3034: feat(tokens): allow using JWT grants from Service Clients

We should be able to just cherry-pick them back, sans the db migrations which will no longer be required.

seanmonstar:
Closes #353


rfk:
adding "shipit" so we remember to revisit this in the train-cutting meeting tomorrow

seanmonstar:
So, the test passes locally, and the `push` tests pass. The `pr` tests seems to have the process exit early, when explicitly testing `process.exit`. I'm going to merge since it seems to be a false positive.

lib/db/mysql/schema.sql
```diff
@@ -41,7 +41,7 @@ CREATE TABLE IF NOT EXISTS tokens (
   FOREIGN KEY (clientId) REFERENCES clients(id) ON DELETE CASCADE,
   userId BINARY(16) NOT NULL,
   INDEX tokens_user_id(userId),
-  email VARCHAR(256) NOT NULL,
+  email VARCHAR(256),
```
rfk: IIUC, this puts `schema.sql` out of sync with the migration patches, which restored the `NOT NULL` constrain on this column.

lib/routes/token.js
```diff
@@ -89,6 +229,69 @@ function confirmRefreshToken(params) {
...
+    if ((payload.exp || -Infinity) < now) {
+      logger.debug('jwt.invalid.exp', { now: now, exp: payload.exp });
+      throw AppError.invalidAssertion();
+    }
+
+    return {
+      clientId: client.id,
+      userId: uid,
+      scope: payload.scope,
+      email: ''
```
rfk: nit: we should probably put a comment on this to explain wtf we're doing returning an empty string :-)

PR #367, opened 2015-12-01
feat(token): reject expired tokens

### LINKED ISSUE #365: Enforce expiry of access tokens
Our access tokens have a lifetime, but it's not currently being enforced because we weren't confident that clients would behave correctly.  Pending final confirmation of #324, we're now confident that existing clients will handle token expiry correctly, so we should bring it back before any *new* clients get a chance to show up and not handle it correctly.

seanmonstar:
Closes #365

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -24,7 +24,7 @@ exports.verify = function verify(token) {
         created_at: token.createdAt,
         expires_at: token.expiresAt
       });
-      //throw AppError.expiredToken(token.expiresAt);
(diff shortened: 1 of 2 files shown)


rfk:
r+; @jrgm would you be comfortable to ship this in train-51 or do you want to hold off until after mozlando?

rfk:
I don't know what travis's problem is tho; some sort of race maybe, since it only fails sporadically?  I don't think there's any way it could be caused by this PR.

PR #368, opened 2015-12-03
feat(openid): add initial OpenID Connect support

### LINKED ISSUE #362: Implement support for `openid` scope and `id_token` response value
To be compatible with OpenID Connect [1], we need to support an `openid` scope and to return an `id_token` JWT response parameter when it is specified by the relier.

[1] http://openid.net/specs/openid-connect-core-1_0.html

seanmonstar:
Requesting authorization with the scope `openid` will result in an
`id_token` property being returned along with the `access_token`.

Closes #362

-----

Known missing:

- [x] `/v1/jwks` route which the openid-configuration document points at

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
 
@@ -152,6 +153,27 @@ const conf = convict({
       default: 10
     }
   },
+  openid: {
+    key: {
+      doc: 'Private JWK to sign id_tokens',
+      default: {}
+    },
+    oldKey: {
+      doc: 'The previous public key that was used to sign id_tokens',
+      default: {}
+    },
+    issuer: {
+      // this should match `issuer` in the 'OpenID Provider Metadata' document
+      // from the fxa-content-server
+      doc: 'The value of the `iss` property of the id_token',
+      default: 'https://accounts.firefox.com'
+    },
+    ttl: {
+      doc: 'Number of milliseconds until id_token should expire',
+      default: '5 minutes',
+      format: 'duration'
+    }
+  },
   publicUrl: {
     format: 'url',
     env: 'PUBLIC_URL',
@@ -220,7 +242,6 @@ conf.validate(options);
 
 // custom validation, since we cant yet specify rules for inside arrays
 conf.get('serviceClients').forEach(function(client) {
-  const assert = require('assert');
   assert(client.id, 'client id required');
   assert.equal(client.id.length, 16, 'client id must be 16 hex digits');
   assert.equal(Buffer(client.id, 'hex').toString('hex'), client.id,
@@ -229,5 +250,23 @@ conf.get('serviceClients').forEach(function(client) {
   assert.equal(typeof client.scope, 'string', 'client scope required');
   assert.equal(typeof client.jku, 'string', 'client jku required');
 });
+
+var key = conf.get('openid.key');
+assert.equal(key.kty, 'RSA', 'openid.key.kty must be RSA');
+assert(key.kid, 'openid.key.kid is required');
+assert(key.n, 'openid.key.n is required');
+assert(key.e, 'openid.key.e is required');
+assert(key.d, 'openid.key.d is required');
+
+var oldKey = conf.get('openid.oldKey');
+if (Object.keys(oldKey).length) {
+  assert.equal(oldKey.kty, 'RSA', 'openid.oldKey.kty must be RSA');
+  assert(oldKey.kid, 'openid.oldKey.kid is required');
+  assert.notEqual(key.kid, oldKey.kid,
+    'openid.key.kid must differ from oldKey');
+  assert(oldKey.n, 'openid.oldKey.n is required');
+  assert(oldKey.e, 'openid.oldKey.e is required');
+  assert(!oldKey.d, 'openid.oldKey.d is forbidden');
+}
 
 module.exports = conf;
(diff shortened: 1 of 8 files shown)


lib/config.js
```diff
@@ -152,6 +153,22 @@ const conf = convict({
       default: 10
     }
   },
+  openid: {
+    key: {
+      doc: 'Private JWK to sign id_tokens',
+      default: {}
+    },
```
rfk: We should get ops/opsec input on the right way to load this key from a secret-management perspective, e.g. in the auth-server we load it from a separate file, which might make some things with permissions easier.

test/api.js
```diff
@@ -1179,6 +1179,38 @@ describe('/v1', function() {
...
+
+          assert.equal(header.alg, 'RS256');
+          assert.equal(header.kid, config.get('openid.key').kid);
+
+          assert.equal(claims.sub, USERID);
+          assert.equal(claims.aud, clientId);
+          assert.equal(claims.iss, config.get('openid.issuer'));
+          var now = Math.floor(Date.now() / 1000);
+          assert(claims.iat <= now);
+          assert(claims.exp > now);
```
rfk: We should check that it also returned an `access_token`, since that will be needed for hitting the UserInfo endpoint

lib/config.js
seanmonstar: @jrgm hi there! how would you prefer we load a secret key for signing JWTs? This currently has it being in JSON form in the config, but is a file better?

vladikoff: cc @jrgm

jrgm: In the current form with json is fine with me.

test/api.js
seanmonstar: Added.

seanmonstar:
@rfk I've added a `/v1/jwks` route.

lib/config.js
```diff
@@ -152,6 +153,27 @@ const conf = convict({
...
+    oldKey: {
+      doc: 'The previous public key that was used to sign id_tokens',
+      default: {}
+    },
```
rfk: :heart:

lib/config.js
```diff
@@ -152,6 +153,27 @@ const conf = convict({
...
+    issuer: {
+      // this should match `issuer` in the 'OpenID Provider Metadata' document
+      // from the fxa-content-server
+      doc: 'The value of the `iss` property of the id_token',
+      default: 'https://accounts.firefox.com'
```
rfk: I wonder if we should de-duplicate this somehow with the `contentUrl` config option.  Or, should the fact that the issuer is the content-server be considered an implementation detail?

seanmonstar: It's not technically required to be the same... I'm +0 on either, really.

lib/routing.js
```diff
@@ -61,6 +61,11 @@ exports.routes = [
     method: 'POST',
     path: v('/verify'),
     config: require('./routes/verify')
+  },
+  {
+    method: 'GET',
+    path: v('/jwks'),
+    config: require('./routes/jwks')
```
rfk: This `/jwks` route needs some tests, in particular:

* Testing that we don't send the private key information as part of this document
* Testing that the oldKey is correctly served when present

lib/config.js
rfk: k, let's just leave it to keep things simple, we can always revisit

lib/routing.js
rfk: Also we should add it to the API doc

lib/routes/jwks.js
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const config = require('../config');
+
+const KEYS = (function() {
+  var priv = config.get('openid.key');
+
+  function pub(key) {
```
rfk: The functionality of this helper is so critical to security, I think it deserves a comment to say "Extract just the public information from the key" and to tell future refactorers to be super careful about it.

lib/routes/token.js
```diff
@@ -29,11 +29,19 @@ const GRANT_JWT = 'urn:ietf:params:oauth:grant-type:jwt-bearer';
...
 }));
 
+const SCOPE_OPENID = 'openid';
+
+const ID_TOKEN_EXPIRATION = Math.floor(config.get('openid.ttl') / 1000);
+const ID_TOKEN_ISSUER = config.get('openid.issuer');
+const ID_TOKEN_KEY = JwTool.JWK.fromObject(config.get('openid.key'), {
+  iss: ID_TOKEN_ISSUER
+});
+
```
rfk: I wonder if we should also say a few words about this in the API docs for this endpoint.  We don't need to explain what OIDC *is*, but we can broadly say that this endpoint is compatible with OIDC code flow or something similar.

rfk:
Otherwise :+1:, this LGTM!

lib/routes/token.js
seanmonstar: Added a comment to the `/authorization` endpoint, and the `/token` endpoint.

lib/routes/jwks.js
seanmonstar: Added warning comment.

lib/routing.js
seanmonstar: Added tests and docs.

lib/routes/jwks.js
rfk: :grin:

rfk:
LGTM!  We likely need to add the new config values in fxa-dev before merge, in order to avoid breaking latest.dev.lcip.org, and a follow-up bug in puppet-config to get it added in production.

rfk:
Putting this in "blocked" since we need to get config changes in before merging

PR #371, opened 2015-12-18
feat(openid): Add support for OIDC `login_hint` query param.

rfk:
@seanmonstar a tiny PR to your PR.  This adds support for the OIDC `login_hint` query param, which I needed for one of the demos at Mozlando.


seanmonstar:
r+

Does this need to merged into the oidc branch? I think it can live on its own, and just merge into master.

PR #372, opened 2015-12-20
chore(awsbox): remove unused awsbox

jrgm:
I think this is unused, so just remove it if so. 

Will also require a related cleanup in fxa-dev which I'll put in a PR for now.

r? - @seanmonstar, @vladikoff


jrgm:
Fxa-dev config change at https://github.com/mozilla/fxa-dev/pull/210

PR #382, opened 2016-01-28
Add support for changing client key with no downtime

### LINKED ISSUE #317: Provide a procedure for changing client secrets without service outage
If we ever need to change the client_secret associated with an oauth client, IIUC there's currently no way to do that without them taking a small service outage.  We have to simultaneously update the secret in the config on our servers and the config on our clients.

We should consider formalizing a pattern for rotating these secrets without a service outage, e.g. by having both a "current secret" and "previous secret" to allow the following sequence:

* generate the new secret on our side, make it the current secret and bump the old secret to "previous"
* send the new secret to the client and wait for them to change over to
* remove the old secret entirely

vbudhram:
This PR adds initial support for changing a client's secret key with no downtime. Adds a new database column `hashSecretPrevious` and updated logic in `/token` route to check if client using previous key.

Fixes #317

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -118,9 +118,9 @@ MysqlStore.connect = function mysqlConnect(options) {
 
 const QUERY_CLIENT_REGISTER =
   'INSERT INTO clients ' +
-  '(id, name, imageUri, hashedSecret, redirectUri, termsUri,' +
+  '(id, name, imageUri, hashedSecret, hashedSecretPrevious, redirectUri, termsUri,' +
   'privacyUri, trusted, canGrant) ' +
-  'VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);';
+  'VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?);';
 const QUERY_CLIENT_DEVELOPER_INSERT =
   'INSERT INTO clientDevelopers ' +
   '(rowId, developerId, clientId) ' +
@@ -149,6 +149,7 @@ const QUERY_CLIENT_LIST = 'SELECT id, name, redirectUri, imageUri, ' +
 const QUERY_CLIENT_UPDATE = 'UPDATE clients SET ' +
   'name=COALESCE(?, name), imageUri=COALESCE(?, imageUri), ' +
   'hashedSecret=COALESCE(?, hashedSecret), ' +
+  'hashedSecretPrevious=COALESCE(?, hashedSecretPrevious), ' +
   'redirectUri=COALESCE(?, redirectUri), ' +
   'termsUri=COALESCE(?, termsUri), privacyUri=COALESCE(?, privacyUri), ' +
   'trusted=COALESCE(?, trusted), canGrant=COALESCE(?, canGrant) ' +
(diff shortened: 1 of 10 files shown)


lib/db/mysql/patches/patch-013-014.sql
```diff
@@ -0,0 +1,7 @@
+-- Re-create NOT NULL constraint on the email column.
+-- We weren't able to drop it in production, this migration
+-- brings it back in our dev environments.
+
+ALTER TABLE tokens ADD COLUMN hashSecretPrevious BINARY(32);
```
vladikoff: should this be `clients` instead of tokens?

vbudhram: Yes, it should :(, will update

GitCop:

 * Commit: 9cfdc1daa8d0a77f0e387b30e3be68f6cc7b9fd9
  * Your subject line is longer than 100 characters

---

lib/db/mysql/index.js
```diff
@@ -218,6 +219,7 @@ MysqlStore.prototype = {
       client.name,
       client.imageUri || '',
       buf(client.hashedSecret),
+      buf(client.hashedSecretPrevious || ''), // TODO: What should default be?
```
seanmonstar: Default should probably be `null`.

lib/db/mysql/patches/patch-013-014.sql
```diff
@@ -0,0 +1,7 @@
+-- Re-create NOT NULL constraint on the email column.
```
seanmonstar: This comment seems peculiar :D

lib/routes/token.js
```diff
@@ -181,11 +181,22 @@ function confirmClient(id, secret) {
...
+          logger.info('client.matchSecretPrevious', { client: id });
+          return client;
+        }
+      }
+
       logger.info('client.mismatchSecret', { client: id });
       logger.verbose('client.mismatchSecret.details', {
         submitted: submitted,
-        db: stored
+        db: storedPrevious ? storedPrevious : stored
```
seanmonstar: I'd change this to include both `stored` and `storedPrevious`. It's at the verbose level, may as well include everything to help debug.

test/api.js
```diff
@@ -703,6 +705,53 @@ describe('/v1', function() {
           assert.equal(res.result.message, 'Incorrect secret');
         }).done(done, done);
       });
+
+      describe('previous secret', function() {
```
seanmonstar: :+1:

lib/routes/token.js
```diff
@@ -181,11 +181,22 @@ function confirmClient(id, secret) {
 
     var submitted = hex(encrypt.hash(buf(secret)));
     var stored = hex(client.hashedSecret);
+
     if (submitted !== stored) {
+      var storedPrevious;
+      if (client.hashedSecretPrevious){
```
seanmonstar: nit: space between parens and brace `if (foo) {`

lib/routes/token.js
```diff
@@ -181,11 +181,22 @@ function confirmClient(id, secret) {
...
+        // Check if secret used is the current previous secret
+        storedPrevious = hex(client.hashedSecretPrevious);
+        if (submitted === storedPrevious){
```
seanmonstar: same nit

seanmonstar:
Nearly perfect! After updates, could you squash into a `feat(clients)` commit. This is a new feature!

GitCop:

 * Commit: 9cfdc1daa8d0a77f0e387b30e3be68f6cc7b9fd9
  * Your subject line is longer than 100 characters

---

vbudhram:
@seanmonstar Nice catches! Made changes, possible to take a quick look over again?

PR #392, opened 2016-02-18
Extract process dot error

### LINKED ISSUE #376: flaky test on travis reporting duplicate database info
For example, see https://api.travis-ci.org/jobs/101651867/log.txt?deansi=true

      2) should not insert already existing clients
fxa-oauth-server.db.mysql.ERROR: updateDbSchema Error: ER_DUP_FIELDNAME: Duplicate column name 'authAt'
    at Query.Sequence._packetToError (/home/travis/build/mozilla/fxa-oauth-server/node_modules/mysql/lib/protocol/sequences/Sequence.js:48:14)

Doesn't always happen. Not sure what's up with it.

rfk:
Fixes #376, by moving calls to `process.exit(1)` out of the code under test and into the top-level scripts.  Instead we throw an error and let the calling code decide what to do about it.

I left the tests running in a tight loop for 10 mins or so, with no failures, so I'm confident this will resolve the flakiness we've been seeing.

@seanmonstar r?


lib/db/mysql/patches/patch-005-006.sql
```diff
@@ -14,7 +14,7 @@ CREATE TABLE refreshTokens (
 
 -- Add expiresAt column for access tokens
 
-ALTER TABLE tokens ADD COLUMN expiresAt TIMESTAMP NOT NULL;
+ALTER TABLE tokens ADD COLUMN expiresAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
rfk: Unrelatedly, this patch was failing to apply on my local mysql, complaining about creating a non-null column with no default value.

seanmonstar: That's because your mysql is a higher version. I believe we run 5.5 in production, and 5.6+ made this breaking change (yay).

seanmonstar:
Yes! This is a far better idea.

... Looks like travis doesn't agree, though. Specifically, your modifying the timestamp column.

rfk:
`"there can be only one TIMESTAMP column with CURRENT_TIMESTAMP in DEFAULT`

Right, of course, that's a perfectly sensible limitation...

rfk:
Hey look, it's our friend "BN_mod_inverse:no inverse", thanks Travis!

lib/db/mysql/patches/patch-005-006.sql
jrgm: I have looked at the specifics here, but production runs 5.6 (and has always run that). Just to be clear on that point.

jrgm: er, s/have/haven't/

PR #394, opened 2016-02-21
Fix a couple of dependency related things

jrgm:
1. `npm-shrinkwrap.json` was dropped. This puts it back.

2. In getting travis tests to run with nodejs 4, generate-rsa-keypair wouldn't compile. So, I forked it, fixed it to use nan@2 and submitted the patch upstream. Now that upstream released that (a while ago), my branch is not needed, so this switches the dependency back to it.

r? @rfk or @seanmonstar ?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,5442 @@
+{
+  "name": "fxa-oauth-server",
+  "version": "0.56.0",
+  "dependencies": {
+    "blanket": {
+      "version": "1.1.6",
+      "from": "blanket@1.1.6",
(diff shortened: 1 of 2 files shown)


jrgm:
Oh, yeah. There's also a problem with oauth on fxa-dev on a brand new box, but I'll fix that this evening.

jrgm:
Actually, this branch fixes the issue with fxa-dev.

rfk:
> npm-shrinkwrap.json was dropped.

Whoops, likely my bad, I must have got distracted half-way through updating it and forgot to re-generate before committing.  Thanks @jrgm!

PR #398, opened 2016-03-07
chore(docs): Add a comment about privKey/pubKey confusion in gen_keys

### LINKED ISSUE #397: Why is scripts/gen_keys saving pubkey for old and privkey for new
https://github.com/mozilla/fxa-oauth-server/blob/master/scripts/gen_keys.js#L38-L48

Is that a typo?

rfk:
Fixes #397.  This is why comments were invented :-)


rfk:
@seanmonstar r?

PR #401, opened 2016-03-11
fix(tokens): Purge expired access tokens

### LINKED ISSUE #366: Purge expired access tokens from the db
Once an access token is expired, there's no point it leaving it hanging around in the db taking up space.  We should set up a background pruning job to get rid of them.

Depends on #365, for obvious reasons.

vbudhram:
This PR adds a stored procedure to remove expired tokens for the oauth database. Procedure can specific max tokens to delete and to ignore specific clientIds.

Fixes #366


lib/db/mysql/scripts/purge-expired-tokens.sql
```diff
@@ -0,0 +1,43 @@
...
+		LEAVE PURGE_LOOP;
+	END IF;
+
+	# Perform a round of expired token deletions
+	SET SQL_SAFE_UPDATES = 0;
+	DELETE p1 FROM `fxa_oauth`.`tokens` AS p1
+	INNER JOIN (
+		SELECT token FROM `fxa_oauth`.`tokens` WHERE expiresAt < NOW() AND UNHEX(clientId) != ignoreClientId ORDER BY expiresAt ASC LIMIT limitDeletePerRound
+	) AS P2
+	USING (token);
```
rfk: MySQL has a non-standard `DELETE FROM table WHERE condition LIMIT x` statement that may be useful to simplify this.

lib/db/mysql/scripts/purge-expired-tokens.sql
```diff
@@ -0,0 +1,43 @@
...
+
+BEGIN
+  DECLARE startIndex INT DEFAULT 1 ;
+  DECLARE endCount INT DEFAULT roundCount ;
+  DECLARE expiredItemCount INT DEFAULT 0;
+
+  PURGE_LOOP: LOOP
+
+	# Exit loop if no expired items found
+    SET expiredItemCount = (SELECT COUNT(*) FROM tokens WHERE expiresAt < NOW() AND UNHEX(clientId) != ignoreClientId ORDER BY expiresAt);
```
rfk: I wonder whether we need to do this as a separate explicit `count`, or whether we can ask MySQL how many rows it successfully deleted in the query below, and exit if it didn't find any rows to delete.

lib/db/mysql/scripts/purge-expired-tokens.sql
rfk: In order to do this efficiently, we'll need to add an index on the `expiresAt` column.  Without it MySQL will have to do a full table scan on every iteration of the loop:

```
mysql> EXPLAIN DELETE p1 FROM `fxa_oauth`.`tokens` AS p1 INNER JOIN (SELECT token FROM `fxa_oauth`.`tokens` WHERE expiresAt < NOW() AND UNHEX(clientId) != '123445' ORDER BY expiresAt ASC LIMIT 100) AS P2;
+----+-------------+------------+------+---------------+------+---------+------+-------+---------------------------------------+
| id | select_type | table      | type | possible_keys | key  | key_len | ref  | rows  | Extra                                 |
+----+-------------+------------+------+---------------+------+---------+------+-------+---------------------------------------+
|  1 | PRIMARY     | <derived2> | ALL  | NULL          | NULL | NULL    | NULL |   100 | NULL                                  |
|  1 | PRIMARY     | p1         | ALL  | NULL          | NULL | NULL    | NULL | 34077 | Using join buffer (Block Nested Loop) |
|  2 | DERIVED     | tokens     | ALL  | NULL          | NULL | NULL    | NULL | 34077 | Using where; Using filesort           |
+----+-------------+------------+------+---------------+------+---------+------+-------+---------------------------------------+
3 rows in set (0.17 sec)
```

The `Using where; Using filesort` here is basically MySQL telling us it's going to have a bad time :-)

The `DELETE LIMIT` form is better as it can avoid sorting all the rows, but `type=ALL` tells us it'll still need a full table scan:

```
mysql> EXPLAIN DELETE  FROM `fxa_oauth`.`tokens` WHERE expiresAt < NOW() AND UNHEX(clientId) != '123445' LIMIT 100;
+----+-------------+--------+------+---------------+------+---------+------+-------+-------------+
| id | select_type | table  | type | possible_keys | key  | key_len | ref  | rows  | Extra       |
+----+-------------+--------+------+---------------+------+---------+------+-------+-------------+
|  1 | SIMPLE      | tokens | ALL  | NULL          | NULL | NULL    | NULL | 34077 | Using where |
+----+-------------+--------+------+---------------+------+---------+------+-------+-------------+
1 row in set (0.00 sec)
```

I can't think of a good way to do this efficiently without adding an index on `expiresAt`.

lib/db/mysql/scripts/purge-expired-tokens.sql
```diff
@@ -0,0 +1,43 @@
...
+    IF expiredItemCount = 0 THEN
+		LEAVE PURGE_LOOP;
+	END IF;
+
+	# Perform a round of expired token deletions
+	SET SQL_SAFE_UPDATES = 0;
+	DELETE p1 FROM `fxa_oauth`.`tokens` AS p1
+	INNER JOIN (
+		SELECT token FROM `fxa_oauth`.`tokens` WHERE expiresAt < NOW() AND UNHEX(clientId) != ignoreClientId ORDER BY expiresAt ASC LIMIT limitDeletePerRound
```
rfk: The clientids are stored as `binary(8)` in the db, so I suspect you'll want this the other way around - passing in `ignoreClientId` as a hex string and doing `clientId != UNHEX(ignoreClientId)` to compare it to the bytes in the db.

lib/db/mysql/scripts/purge-expired-tokens.sql
rfk: (You can also do `DELETE ... ORDER BY ... LIMIT` if you want to ensure it deletes the oldest tokens first, but I don't think that's necessary.  We want to start rejecting all expired tokens at the app layer, so it won't matter what order they're deleted from the database).

lib/db/mysql/scripts/purge-expired-tokens.sql
```diff
@@ -0,0 +1,43 @@
...
+  PURGE_LOOP: LOOP
+
+	# Exit loop if no expired items found
+    SET expiredItemCount = (SELECT COUNT(*) FROM tokens WHERE expiresAt < NOW() AND UNHEX(clientId) != ignoreClientId ORDER BY expiresAt);
+    IF expiredItemCount = 0 THEN
+		LEAVE PURGE_LOOP;
+	END IF;
+
+	# Perform a round of expired token deletions
+	SET SQL_SAFE_UPDATES = 0;
```
rfk: Did you get an error that caused you to add this, or is it just for completeness?  I don't *think* we're running in safe-update mode in production but I could easily be mistaken.

lib/db/mysql/scripts/purge-expired-tokens.sql
rfk: Aha, `ROW_COUNT()` is the one I was thinking of: http://dev.mysql.com/doc/refman/5.0/en/information-functions.html#function_row-count

rfk:
Today I learned, that MySQL lets you explicit request a non-locking alter table.  I tried doing this:

```
ALTER TABLE tokens ADD INDEX idx_expiresAt (expiresAt), ALGORITHM = INPLACE, LOCK = NONE;
```

It didn't complain, which I assume means that it was happy to add the index in a non-locking fashion.  The resulting delete successfully uses the index:

```
mysql> EXPLAIN DELETE  FROM `fxa_oauth`.`tokens` WHERE expiresAt < NOW() AND UNHEX(clientId) != '123445' LIMIT 100;
+----+-------------+--------+-------+---------------+---------------+---------+-------+-------+-------------+
| id | select_type | table  | type  | possible_keys | key           | key_len | ref   | rows  | Extra       |
+----+-------------+--------+-------+---------------+---------------+---------+-------+-------+-------------+
|  1 | SIMPLE      | tokens | range | idx_expiresAt | idx_expiresAt | 4       | const | 17038 | Using where |
+----+-------------+--------+-------+---------------+---------------+---------+-------+-------+-------------+
1 row in set (0.00 sec)
```

rfk:
@jrgm does this `LOCK=NONE` business I discovered above work out in production?  If so then ISTM we should start specifying it explicitly in our database migrations etc.

lib/db/mysql/scripts/purge-expired-tokens.sql
vbudhram: Nice! I didn't know it had that function, will update to it

rfk:
From a code-organization standpoint, if we're going to use a stored procedure for this, I think it still makes sense to maintain it alongside the other schema and migration files in the repo rather than in a separate directory.  I'm thinking something like:

* Add a new migration to `lib/db/mysql/patches/` that creates the procedure and adds the index, and a reverse migration to drop them
* Update `lib/db/mysql/schema.sql` to reflect the existence of the index and stored procedure, so that it remains the canonical overview of the desired state of the db

The helper script, if you plan to keep it, can live separately because I don't think we want it in production.

lib/db/mysql/scripts/purge-expired-tokens.sql
vbudhram: Yea I was getting that error about it being an unsafe update. I will follow up with @jrgm on how production and stage environments are setup.

lib/db/mysql/scripts/purge-expired-tokens.sql
```diff
@@ -0,0 +1,43 @@
+DROP PROCEDURE IF EXISTS PURGE_EXPIRED_TOKENS;
+
+DELIMITER $$
+# Create a procedure that will delete a total of `roundCount` * `limitDeletePerRound` items, every `delaySeconds`.
+# Skip tokens that belong clientId
+CREATE PROCEDURE PURGE_EXPIRED_TOKENS(IN roundCount INT, IN limitDeletePerRound INT, IN delaySeconds INT, IN ignoreClientId CHAR(10))
```
rfk: API usability nit: it may be better to let the caller pass in the desired maximum number of tokens to delete, rather than the number of rounds to execute.  This seems like the thing that the caller is most likely to want to control, so we shouldn't require them to calculate the values of `roundCount` and `limitDeletePerRound` that would achieve it.

lib/db/mysql/scripts/purge-expired-tokens.sql
vbudhram: I like that idea. As a side note, do you think it makes more sense for the procedure to delete tokens from a specific clientId instead of all clients but ignoring a single clientId? Seems like we could have a little more flexibility this way.

rfk: I don't think we'd use that in practice - this special case with pocket is a one-off and hopefully will be short-lived.

lib/db/mysql/patches/patch-014-015.sql
```diff
@@ -0,0 +1,50 @@
+-- Add purge expired tokens stored procedure and index to expiresAt column
+ALTER TABLE tokens ADD INDEX idx_expiresAt (expiresAt);
```
vbudhram: When using Travis, `ALTER TABLE tokens ADD INDEX idx_expiresAt (expiresAt), ALGORITHM = INPLACE, LOCK = NONE;` causes the tests to fail with 

`fxa-oauth-server.db.mysql.ERROR: updateDbSchema Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ALGORITHM = INPLACE, LOCK = NONE;`

I think this might be related to the version of mysql running in Travis because it passes locally with 5.6.

@jrgm do you know which version of mysql is running in aws?

vbudhram:
@rfk Made updates, mind another review?

lib/db/mysql/patches/patch-014-015.sql
vladikoff: @vbudhram 5.6.21 in prod : https://github.com/mozilla/fxa-auth-db-mysql#prerequisites

You can switch travis to 5.6 : https://docs.travis-ci.com/user/database-setup/#MySQL-5.6

vbudhram: Doh! Would have been helpful if I read through the docs. I'll update travis config and try again. Thanks @vladikoff !

.travis.yml
```diff
@@ -5,7 +5,8 @@ node_js:
   - '0.12'
   - '4'
 
-sudo: false
+dist: trusty
+sudo: true
```
vladikoff: are these needed? `sudo: true` slows thing down, 

is it possible to just add 

```
     - mysql-server-5.6
     - mysql-client-core-5.6
     - mysql-client-5.6
```

vbudhram: @vladikoff looks like it needs to have `sudo: true` or else it fails with https://travis-ci.org/mozilla/fxa-oauth-server/jobs/118151093

lib/db/mysql/schema.sql
```diff
@@ -46,7 +46,8 @@ CREATE TABLE IF NOT EXISTS tokens (
   type VARCHAR(16) NOT NULL,
   scope VARCHAR(256) NOT NULL,
   createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  expiresAt TIMESTAMP NOT NULL
+  expiresAt TIMESTAMP NOT NULL,
+  INDEX idx_expiresAt(expiresAt),
```
vbudhram: > Update lib/db/mysql/schema.sql to reflect the existence of the index and stored procedure, so that it remains the canonical overview of the desired state of the db

@rfk Not sure what you meant by showing existence of stored procedure? Did you want to have `CREATE PROCEDURE PURGE_EXPIRED_TOKENS` statement here? Or did you want to have the entire stored procedure?

vbudhram:
@vladikoff @jrgm Forgot Ryan is out, would you guys mind doing an r?

lib/db/mysql/schema.sql
rfk: > Did you want to have CREATE PROCEDURE PURGE_EXPIRED_TOKENS statement here

Yes, the entire create statement; ideally you could run this file and put your database in the correct "current" state without worrying about migrations etc.  It's an unfortunate bit of duplication but having the ability to see the whole schema in a single file is a real benefit.

lib/db/mysql/index.js
```diff
@@ -460,6 +460,16 @@ MysqlStore.prototype = {
     });
   },
 
+  purgeExpiredTokens: function purgeExpiredTokens(numberOfTokens, deplaySeconds, ignoreClientId){
```
rfk: "deplay"?

rfk:
This is looking good, I like that it's getting simpler and more testable as we go.

One thought, how straightforward would it be to add a convenience script `./bin/purge_expired_tokens.js` that loads the config, instantiates the db, and does a call to the purge method?  This might be a nice affordance for ops as we initially roll this out to production.

test/db/index.js
```diff
@@ -94,6 +96,109 @@ describe('db', function() {
...
+          .then( function () {
+            return seedTokens(clientIdA, userId, email, 500, new Date(Date.now() - (1000 * 600)));
+          })
+          .then( function () {
+            return seedTokens(clientIdB, userId, email, 500, new Date(Date.now() - (1000 * 600)));
+          });
+      });
+
+      it('should call purgeExpiredTokens', function() {
+        return db.purgeExpiredTokens(1000, 5)
```
rfk: It's pretty important that, in production, we don't accidentally call this procedure with a `NULL` ignoreClientId.  I wonder if we can make it more robust by e.g. reading the value from config rather than requiring the caller to provide it every time.

vbudhram:
@rfk, added the last remaining bits to this PR, a convenience script, requiring an `ignoreClientId` to run and updated docs. Review at your leisure :)

bin/purge_expired_tokens.js
```diff
@@ -0,0 +1,84 @@
...
+      .then(function(){
+        logger.info('Purge completed!');
+        process.exit(0);
+      })
+      .catch(function(err){
+        logger.error(err);
+        process.exit(1);
+      });
+  } else {
+    logger.debug('Unable to run purge expired tokens, only avalible when using config with mysql database.');
```
rfk: grammar nit: s/Unable to run purge/Unable to purge/

lib/db/mysql/index.js
```diff
@@ -460,6 +460,30 @@ MysqlStore.prototype = {
...
+        ]);
+      })
+      .then(function() {
+        logger.debug('purgeExpiredTokens completed');
+      })
+      .catch(function(err){
+        err = new Error('Invalid ignoreClientId, please ensure client exists.');
+        logger.error(err);
+        throw err;
+      });
```
rfk: nit: in order to void catching error that aren't related to the ignoreClientId, I think this `catch` should be further up in the call chain, e.g.:

```
return getClientDevelopers(ignoreClientId)
  .catch( /*  catch and re-throw errors raised by that specific call */ )
  .then( /* then continue with the other processing stuff */ )
```

bin/purge_expired_tokens.js
```diff
@@ -0,0 +1,84 @@
...
+const ignorePocketClientId = program.pocketId;
+
+db.ping().done(function() {
+
+  // Only mysql impl supports token deletion at the moment
+  if (db.purgeExpiredTokens) {
+
+    // To reduce the risk of deleting pocket tokens, ensure that the pocket-id passed in
+    // belongs to a client.
+    db.getClient(ignorePocketClientId)
```
rfk: IIUC you check this twice, once in the script here and then again in the `db.purgeExpiredTokens` method.  It probably only needs to be done in one place.

lib/db/mysql/patches/patch-014-015.sql
```diff
@@ -0,0 +1,50 @@
...
+# -- (IN) delaySeconds : After every 1000 deletes, delay the next delete by this
+# -- (IN) ignoreClientId: Skip access tokens from this client
+# -- (OUT) totalDeleted: Number deleted tokens
+# -- (OUT) rounds: Number to iterations taken to delete tokens
+#
+#   Ex.
+#       CALL PURGE_EXPIRED_TOKENS(1000,0,'37e7061dd45df02f',@totalDeleted,@rounds);
+#       SELECT @totalDeleted, @rounds;
+#
+CREATE PROCEDURE PURGE_EXPIRED_TOKENS(IN deleteTokenCount INT, IN delaySeconds INT, IN ignoreClientId CHAR(16), OUT totalDeleted INT, OUT rounds INT)
```
rfk: So...the more I learn about stored procedures, the less I feel like adding more of them to our codebase.  My latest gripe:  they apparently operate with the strict-mode setting in force at the time they were created, not the strict-mode in force at the time they're invoked.

How would you feel above moving this looping logic directly into javascript in the `db.purgeExpiredTokens` method?

test/db/index.js
```diff
@@ -94,6 +96,106 @@ describe('db', function() {
     });
   });
 
+  if (config.get('db.driver') === 'mysql') {
```
rfk: :heart: testability!

vbudhram:
Updated this PR to not use stored procedures, review at leisure @rfk

bin/purge_expired_tokens.js
```diff
@@ -0,0 +1,76 @@
...
+const db = require('../lib/db');
+const logger = require('../lib/logging')('bin.purge_expired_tokens');
+
+if (!program.pocketId) {
+  logger.debug('Required pocket client id!');
+  process.exit(1);
+}
+
+const numberOfTokens = parseInt(program.tokenCount) || 200;
+const delaySeconds = parseInt(program.delaySeconds) || 1; // Default 60 seconds
```
rfk: Comment doesn't match value in code

bin/purge_expired_tokens.js
```diff
@@ -0,0 +1,76 @@
...
+const ignorePocketClientId = program.pocketId;
+
+db.ping().done(function() {
+
+  // Only mysql impl supports token deletion at the moment
+  if (db.purgeExpiredTokens) {
```
rfk: It's probably worth erroring out rather than silently doing nothing if `purgeExpiredTokens` method doesn't exist.

rfk: Never mind me, I see you log an error message further down, 👍

test/db/index.js
```diff
@@ -94,6 +96,164 @@ describe('db', function() {
...
+          .catch( function (error) {
+            assert.equal(error.message, 'Invalid ignoreClientId, please ensure client exists.');
+          });
+      });
+
+      it('should call purgeExpiredTokens and ignore client', function() {
+        return db.purgeExpiredTokens(1000, 0, clientIdA)
+          .then( function () {
+            // Check clientA tokens not deleted
+            return db._read('SELECT COUNT(*) FROM fxa_oauth.tokens WHERE clientId=UNHEX(?);', [
```
rfk: nit: if you made this `SELECT COUNT(*) AS count ..` then the part where you read the result might be slightly cleaner, `result[0].count` rather than `result[0]['COUNT(*)']`.

rfk:
Although we've lost the stored procedure, I think we still need a db migration to add the index described in https://github.com/mozilla/fxa-oauth-server/pull/401#issuecomment-200034586

rfk:
Otherwise 👍 looking forward to expiring our giant backlog of tokens 😁

vbudhram:
Added expiresAt index, mind another r? @rfk

bin/purge_expired_tokens.js
rfk: This comment still doesn't match the value used in the code

rfk:
Besides that one docs nit, r+!

PR #402, opened 2016-05-02
chore(travis): drop node 0.12 support

vladikoff:


vladikoff:
@jrgm @seanmonstar r? 

We dropped 0.12 support intravis in the content server / auth server.

seanmonstar:
If we're not going to upgrade to 0.12, seems fine to me. @jrgm?

PR #404, opened 2016-07-15
fix(deps): update some dependencies

vladikoff:


vladikoff:
@vbudhram r?

Should be less surprises here comparing to the auth-server

PR #405, opened 2016-07-16
feat(oauth): add methods to support oauth client management

### LINKED ISSUE #409: Add methods to support oauth client management
Need the following:

* Get clients for UID with non-expired tokens
* Delete all non-expired tokens for some given client id

vladikoff:
Fixes #409
Feature card: https://github.com/mozilla/fxa/pull/171


lib/db/memory.js
```diff
@@ -242,6 +242,18 @@ MemoryStore.prototype = {
     delete this.tokens[unbuf(id)];
     return P.resolve();
   },
+  /**
+   * Get all services that have have non-expired tokens
+   * @param email User's email
+   * @returns {*}
+   */
+  getActiveServicesByEmail: function getActiveServicesByEmail(email) {
```
rfk: In particular, @seanmonstar, I have a feeling we were trying to move away from using email addresses in this codebase, and instead operating using only uids and looking up emails in the auth-server when necessary.  It's been a while, am I remembering that correctly?

vladikoff:
No need for a full review here yet, just quick prototype stuff

lib/db/memory.js
seanmonstar: Correct, since the email address for a user can change, it should always be via `uid`.

lib/db/mysql/index.js
```diff
@@ -177,6 +177,9 @@ const QUERY_CODE_DELETE_USER = 'DELETE FROM codes WHERE userId=?';
 const QUERY_DEVELOPER = 'SELECT * FROM developers WHERE email=?';
 const QUERY_DEVELOPER_DELETE = 'DELETE FROM developers WHERE email=?';
 const QUERY_PURGE_EXPIRED_TOKENS = 'DELETE FROM tokens WHERE clientId != UNHEX(?) AND expiresAt < NOW() LIMIT ?;';
+// TODO: better names and only active tokens, not all.
+const QUERY_ACTIVE_SERVICES_FOR_EMAIL = 'SELECT c.name, c.id, t.scope FROM fxa_oauth.clients AS c, fxa_oauth.tokens as t WHERE c.id = t.clientId AND email = ?';
+const QUERY_DELETE_TOKENS_BY_SERVICE = 'DELETE FROM tokens WHERE clientId=?';
```
seanmonstar: Since they are called `clients` in the codebase, I feel it would probably be less confusing to always call them `client`. So, `s/service/client/gi`?

lib/db/mysql/index.js
```diff
@@ -393,19 +396,54 @@ MysqlStore.prototype = {
...
   removeAccessToken: function removeAccessToken(id) {
     return this._write(QUERY_ACCESS_TOKEN_DELETE, [buf(id)]);
   },
 
+  /**
+   * Get all services that have have non-expired tokens
+   * @param email User's email
+   * @returns {*}
+   */
+  getActiveServicesByEmail: function getActiveServicesByEmail(email) {
```
seanmonstar: `uid` treatment here too.

lib/routes/services/delete.js
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
seanmonstar: https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1destroy

vladikoff: Sorry, what do you mean by this comment?

It seems easier to delete all tokens for some clientId + uid, then call delete for each token.

vladikoff: Oh maybe you mean `delete` -> `destroy` ?

vladikoff:
@seanmonstar could I please get some feedback on this again? 

The functionality is all there, I'm adding more tests.

lib/auth_bearer.js
```diff
@@ -0,0 +1,61 @@
...
+        return reply(AppError.unauthorized('Bearer token not provided'));
+      }
+      var tok = auth.split(' ')[1];
+
+      if (!validators.HEX_STRING.test(tok)) {
+        return reply(AppError.unauthorized('Illegal Bearer token'));
+      }
+
+      token.verify(tok).done(function tokenFound(details) {
+        if (! hasAuthScope(details.scope)) {
```
seanmonstar: This can be done via a config on the route.

lib/routes/client-tokens/delete.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const db = require('../../db');
+
+module.exports = {
+  auth: {
+    strategy: 'authBearer'
```
seanmonstar: add a `scope: ['clients:write']` here. Or, bonus points for importing the constant.

lib/routes/client-tokens/list.js
```diff
@@ -0,0 +1,37 @@
...
+    name: client.name,
+    id: hex(client.id),
+    lastAccessTime: lastAccessTime,
+    lastAccessTimeFormatted: lastAccessTimeFormatted
+  };
+}
+
+module.exports = {
+  auth: {
+    strategy: 'authBearer'
```
seanmonstar: `scope: ['clients:write']`

lib/auth_bearer.js
seanmonstar: Example of it in profile server: https://github.com/mozilla/fxa-profile-server/blob/master/lib/routes/email.js#L17

lib/db/mysql/index.js
```diff
@@ -388,19 +398,62 @@ MysqlStore.prototype = {
...
+      return P.reject(new Error('Uid is required'));
+    }
+
+    return this._read(QUERY_ACTIVE_CLIENT_TOKENS_BY_UID, [
+      buf(uid)
+    ]);
+  },
+
+  /**
+   * Delete all non-expired tokens for some clientId and uid.
```
seanmonstar: non-expired part isn't entirely true. This query deletes **all** tokens for this clientId/uid pair.

Not that it shouldn't, just that the comment isn't in sync.

lib/db/mysql/index.js
```diff
@@ -388,19 +398,62 @@ MysqlStore.prototype = {
...
+   *
+   * @param {String} clientId Client ID
+   * @param {String} uid User Id as Hex
+   * @returns {Promise}
+   */
+  deleteActiveClientTokens: function deleteActiveClientTokens(clientId, uid) {
+    if (! clientId || ! uid) {
+      return P.reject(new Error('clientId and uid are required'));
```
seanmonstar: These error strings worried me (instead of using a proper error code, but then I remembered that none of the other methods do these kinds of checks. Are they required? I believe if the values are null, the query will search for those columns that are null. If the columns are `NOT NULL`, it should always return an result.

test/auth_bearer.js
```diff
@@ -0,0 +1,100 @@
...
+      })
+    });
+
+    it('errors if token has wrong scope', function(done) {
+      var mocks = {
+        './token': {
+          verify: function () {
+            return P.resolve({
+              // no space in scope below
+              scope: ['bar:foo:clients:write'],
```
seanmonstar: I'd add a few other scopes that might seem like they could work, but shouldn't.

Examples: 

- `clients`
- `clients:write:foo`
- `clients:writ`

seanmonstar:
An excellent set of tests, woo!

vladikoff:
@seanmonstar please **r** for merge?

seanmonstar:
LGTM!

Something about the routes named `/client-tokens` and such sticks out at me, but I don't have better suggestions, so shrug.

PR #411, opened 2016-08-16
fix(log): avoid crashing on bad payload

### LINKED ISSUE #410: Check for valid payload when logging.
See this line: https://github.com/mozilla/fxa-oauth-server/blob/master/lib/logging/summary.js#L29

`payload` is possible not an object and will crash the logging code.

vladikoff:
Fixes #410

### DIFF
--- a/lib/logging/summary.js
+++ b/lib/logging/summary.js
@@ -4,7 +4,20 @@
 
 const logger = require('./')('summary');
 
+function parsePayload(payload) {
+  var payloadKeys = ['INVALID_PAYLOAD_OBJECT'];
+  try {
+    // given payload object might not be a valid object
+    // See issue #410
+    payloadKeys = Object.keys(payload);
+  } catch (e) {
+    // failed to parse payload keys.
+  }
+  return payloadKeys;
+}
+
 module.exports = function summary(request, response) {
+  /*eslint complexity: [2, 11] */
   if (request.method === 'options') {
     return;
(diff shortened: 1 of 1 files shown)


vladikoff:
@jrgm please r?

vladikoff:
@jrgm we already have a check for `undefined` or `null` payload so it's probably something that cause then, that's why this uses a `try`

lib/logging/summary.js
```diff
@@ -17,6 +18,15 @@ module.exports = function summary(request, response) {
     scope: request.auth.credentials.scope
   };
 
+  var payloadKeys = [];
+  try {
+    // given payload object might not be a valid object
+    // See issue #410
+    payloadKeys = Object.keys(payload);
+  } catch (e) {
+    // failed to parse payload keys.
```
rfk: I wonder if it's worth adding some sort of sentinel value, like `payloadKeys = ['INVALID_PAYLOAD_OBJECT']` or similar, so that we can easily identify any such cases in the logs.

vladikoff: yeah!

jrgm: Small suggestion. Factor the try/catch into a helper function. Otherwise v8 will not optimize the exported function.

jrgm:
Otherwise, plus my suggestion, and @rfk's suggestion, r+ from me.

PR #414, opened 2016-08-23
fix(test): encrypt refresh_token on db query

### LINKED ISSUE #413: Query encrypted refreshToken.token
When refreshTokens are generated token is [encrypted ](https://github.com/mozilla/fxa-oauth-server/blob/master/lib/db/mysql/index.js#L412). When queried it's not. [here](https://github.com/mozilla/fxa-oauth-server/blob/master/lib/db/mysql/index.js#L426) and [here](https://github.com/mozilla/fxa-oauth-server/blob/master/test/db/index.js#L315).

larissagaulia:
fix #413

### DIFF
--- a/test/db/index.js
+++ b/test/db/index.js
@@ -8,6 +8,7 @@ const assert = require('insist');
 const buf = require('buf').hex;
 const hex = require('buf').to.hex;
 
+const encrypt = require('../../lib/encrypt');
 const db = require('../../lib/db');
 const config = require('../../lib/config');
 const auth = require('../../lib/auth');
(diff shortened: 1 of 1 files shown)


vladikoff:
@seanmonstar r?

test/db/index.js
```diff
@@ -312,7 +321,7 @@ describe('db', function() {
         return db.getAccessToken(token);
       }).then(function (t) {
         assert.equal(t, undefined, 'token deleted');
-        return db.getRefreshToken(refreshToken);
+        return db.getRefreshToken(encrypt.hash(refreshToken));
       }).then(function (t) {
         assert.equal(t, undefined, 'refresh_token deleted');
```
vladikoff: @seanmonstar without the change above this is always undefined, even if the user was not deleted.

seanmonstar:
r+, travis still yellow

vladikoff:
@larissagaulia thanks!! 👍

PR #418, opened 2016-09-12
tuneups for bin/purge_expired_tokens.js

jrgm:
This tunes up bin/purge_expired_tokens.js:

1) log uncaughtException; minimum log level of info
   - moves to more conventional logging in this repo and nothing below log level info
   - prints out any uncaughtException

2) use db.getClient() to check for unknown clientId
   - the `clients` table is the one that matters

3) set db.autoUpdateClients config to false
   - No point in doing this in dev for the purposes of this script, and disabled in prod/stage.

4) moar logging
   - because https://youtu.be/-fQGPZTECYs?t=24

5) Promise.delay takes milliseconds; also allow subsecond delay for command line option
   - nuff said

r? - @vbudhram 

Maybe look at each commit separately, instead of the aggregate.

### DIFF
--- a/bin/purge_expired_tokens.js
+++ b/bin/purge_expired_tokens.js
@@ -47,7 +47,7 @@ if (!program.pocketId) {
 }
 
 const numberOfTokens = parseInt(program.tokenCount) || 200;
-const delaySeconds = parseInt(program.delaySeconds) || 1; // Default 1 seconds
+const delaySeconds = Number(program.delaySeconds) || 1; // Default 1 seconds
 const ignorePocketClientId = program.pocketId;
 
 db.ping().done(function() {
--- a/bin/purge_expired_tokens.js
+++ b/bin/purge_expired_tokens.js
@@ -38,7 +38,7 @@ const db = require('../lib/db');
 const logger = require('../lib/logging')('bin.purge_expired_tokens');
 
 if (!program.pocketId) {
-  logger.debug('Required pocket client id!');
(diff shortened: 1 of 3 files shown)


bin/purge_expired_tokens.js
```diff
@@ -38,39 +42,45 @@ const db = require('../lib/db');
...
+    logger.info('skipping', { message: message });
   }
 }, function(err) {
   logger.critical('db.ping', err);
   process.exit(1);
 });
 
-process.on('uncaughtException', function () {
+process.on('uncaughtException', function (err) {
+  logger.error('error', err);
```
vbudhram: 👍

lib/db/mysql/index.js
```diff
@@ -514,39 +514,53 @@ MysqlStore.prototype = {
...
               if (res.affectedRows === 0) {
+                message = '0 affectedRows. Bailing out.';
+                logger.info('purgeExpiredTokens', { message: message });
                 return;
               }
 
               deletedItems = deletedItems + res.affectedRows;
 
-              return P.delay(delaySeconds)
+              return P.delay(delaySeconds * 1000)
```
vbudhram: Did you intended to multiple by 1000?

jrgm: `P.delay` takes `ms` as an argument.

vbudhram: Ah, good catch!

vbudhram:
@jrgm Thanks for adding these! Changes look good, r+. Merge at will.

PR #419, opened 2016-09-13
fix(purge-expired): accept a list of pocket-id's

jrgm:
r? - @vbudhram 

There is more than one ID that needs to be preserved, so this makes it possible.

### DIFF
--- a/bin/purge_expired_tokens.js
+++ b/bin/purge_expired_tokens.js
@@ -13,6 +13,10 @@
  *
  * node purge_expired_tokens.js --config dev --pocket-id dcdb5ae7add825d2 --token-count 10000 --delay-seconds 1
  *
+ * or, for multiple pocket ids:
+ *
+ * node purge_expired_tokens.js --config dev --pocket-id dcdb5ae7add825d2,678f75ae1c0f8002 --token-count 10000 --delay-seconds 1
+ *
  * */
 
 const config = require('../lib/config');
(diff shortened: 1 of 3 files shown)


bin/purge_expired_tokens.js
```diff
@@ -25,7 +25,7 @@ config.set('db.autoUpdateClients', false);
 program
   .version(package.version)
   .option('-c, --config [config]', 'Configuration to use. Ex. dev')
-  .option('-p, --pocket-id <pocketId>', 'Pocket Client Id. These tokens will not be purged.')
+  .option('-p, --pocket-id <pocketId>', 'Pocket Client Ids. These tokens will not be purged. (CSV)')
```
vbudhram: nit: Could update example command line in https://github.com/mozilla/fxa-oauth-server/pull/419/files#diff-ad4b135c187bb7e96c8db4aa28dfa456L14

lib/db/mysql/index.js
```diff
@@ -518,13 +518,23 @@ MysqlStore.prototype = {
       throw new Error('empty ignoreClientId');
     }
 
-    return self.getClient(ignoreClientId)
-      .then(function (ignoreClient) {
+    if (! Array.isArray(ignoreClientId)) {
+      ignoreClientId = [ ignoreClientId ];
+    }
+
+    var clientIds = ignoreClientId.map(function (id) {
```
vbudhram: Not sure how much ES6 you would like, but you could use arrow functions here.

vbudhram:
@jrgm Didn't know that there were multiple pocket ids. This addition looks good to me 👍, r+, merge at will!

lib/db/mysql/index.js
jrgm: ... if we ran on es6, but not yet.

PR #427, opened 2016-11-01
fix(server): starts internal server [WIP]

### LINKED ISSUE #233: 'npm start' does not start the internal server
Seems like only the main server is started

RPraneetha:
fixes #233

### DIFF
--- a/grunttasks/server.js
+++ b/grunttasks/server.js
@@ -7,7 +7,7 @@ module.exports = function (grunt) {
 
   grunt.config('nodemon', {
     dev: {
-      script: 'bin/server.js',
+      script: 'bin/server.js','bin/internal.js',
       options: {
         args: ['--node-env=dev']
       }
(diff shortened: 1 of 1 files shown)


rfk:
@seanmonstar r?

grunttasks/server.js
```diff
@@ -7,7 +7,7 @@ module.exports = function (grunt) {
 
   grunt.config('nodemon', {
     dev: {
-      script: 'bin/server.js',
+      script: 'bin/server.js','bin/internal.js',
```
seanmonstar: Does this actually parse? This is adding the `'bin/internal.js'` as an additional property of the `dev` object, not as part of `script`. Should this be wrapped in an array? `script: ['bin/server.js', 'bin/internal.js']`?

vladikoff:
Closing this for now, please reopen if you push new changes / code to this.

PR #428, opened 2016-11-09
feat(authorization): add uri validation on the authorization endpoint

### LINKED ISSUE #387: Update Joi version to support uri validation
The new version of Joi supports [uri validation](https://github.com/hapijs/joi/blob/v7.2.3/API.md#stringurioptions). Could use this for `redirectUri` validation

https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/authorization.js#L104

### LINKED ISSUE #388: restrict redirect_uri to `https://` and `urn:`
This extends from [this comment in the content server](https://github.com/mozilla/fxa-content-server/pull/3469#discussion_r52017618). @vladikoff mentioned the OAuth server allows any arbitrary string as the redirect_uri in the call to `/authorization`, which is exceptionally lenient. @rfk and I talked today about tightening this up to only allow `https://` and `urn:` URIs, and additional protocols can be added if required.

I imagine both the `redirect_uri` passed to `/authorization` and the `redirect_uri` registered for each client would both need to be validated.

Ref https://github.com/mozilla/fxa-content-server/pull/3469, #387

vladikoff:
Fixes #387
Fixes #388 

@vbudhram r?

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -95,5 +95,6 @@
   "openid": {
     "keyFile": "../config/key.json",
     "key": {}
-  }
+  },
+  "fxaEnv": "development"
 }
--- a/config/dev.json
+++ b/config/dev.json
@@ -96,5 +96,5 @@
     "keyFile": "../config/key.json",
     "key": {}
   },
-  "fxaEnv": "development"
+  "allowHttpRedirects": true
 }
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -30,6 +30,17 @@ const UNTRUSTED_CLIENT_ALLOWED_SCOPES = [
   'profile:display_name'
 ];
 
+const fxaEnv = config.get('fxaEnv');
+
+var ALLOWED_SCHEMES = [
+  'https'
+];
+
+if (fxaEnv === 'development') {
+  // http scheme used when developing OAuth clients
+  ALLOWED_SCHEMES.push('http');
+}
+
 function isLocalHost(url) {
   var host = new URI(url).hostname();
   return host === 'localhost' || host === '127.0.0.1';
@@ -102,7 +113,11 @@ module.exports = {
       assertion: validators.assertion
         .required(),
       redirect_uri: Joi.string()
-        .max(256),
+        .max(256)
+        // uri validation ref: https://github.com/hapijs/joi/blob/master/API.md#stringurioptions
+        .uri({
+          scheme: ALLOWED_SCHEMES
+        }),
       scope: validators.scope,
       response_type: Joi.string()
         .valid(CODE, TOKEN)
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -30,13 +30,13 @@ const UNTRUSTED_CLIENT_ALLOWED_SCOPES = [
   'profile:display_name'
 ];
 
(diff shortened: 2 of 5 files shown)


seanmonstar:
This doesn't cause problems with `urn:ietf:wg:oauth:2.0:fx:webchannel` that Firefox uses?

vladikoff:
> This doesn't cause problems with urn:ietf:wg:oauth:2.0:fx:webchannel that Firefox uses?

good question, I think only Hello used it and we should not have anymore OAuth reliers that use that. 

Ref https://github.com/mozilla/fxa-content-server/commit/c07664f#diff-f243a8eaa90957e9b805f13009c69169

vladikoff:
This PR also addresses https://github.com/mozilla/fxa-oauth-server/issues/388 somewhat. 

Should we allow `http` here?

vladikoff:
@rfk  ^ ?

vbudhram:
@seanmonstar Yea, there used be to validation for URN redirects previously but it was removed when support for oauth webchannel keys was dropped, https://github.com/mozilla/fxa-content-server/pull/4295

rfk:
> This doesn't cause problems with urn:ietf:wg:oauth:2.0:fx:webchannel that Firefox uses?

Just to confirm, yes, we deliberately removed this as we don't have any webchannel oauth reliers anymore.  We can bring it back later if we need it.

> Should we allow http here? Or should http be only for latest and stable, but not stage and prod? 

Any chance we can restrict the http URLs to localhost only?  Otherwise, I'd be happy with the ability to pref off support for http URLs in production config.  Production reliers should absolutely not be using http redirect URIs (with the possible exception of localhost).

lib/routes/authorization.js
```diff
@@ -102,7 +102,14 @@ module.exports = {
       assertion: validators.assertion
         .required(),
       redirect_uri: Joi.string()
-        .max(256),
+      .max(256)
```
seanmonstar: I agree that consistency in style should be kept.

vladikoff:
@rfk and @vladikoff to discuss this...

rfk:
Whoops, forgot to discuss this in 1:1.  It sounds like the simplest option will be to allow http:// URLs on dev boxes, but only https:// in stage on prod.  @vladikoff thoughts?

vladikoff:
> Whoops, forgot to discuss this in 1:1. It sounds like the simplest option will be to allow http:// URLs on dev boxes, but only https:// in stage on prod. @vladikoff thoughts?

Yes, that's a good plan

vladikoff:
@seanmonstar updated!

vladikoff:
Filed https://github.com/mozilla/fxa-dev/pull/288 to allow `http` clients in fxa-dev

config/dev.json
```diff
@@ -95,5 +95,6 @@
   "openid": {
     "keyFile": "../config/key.json",
     "key": {}
-  }
+  },
+  "fxaEnv": "development"
```
vladikoff: `env` is node environment, fxaEnv is Firefox Accounts running mode. 

We discussed adding this in the future to stop relying on node's env. @jrgm thoughts?

vladikoff: from meeting: let's make a new env variable for this feature and enable it in fxa-dev.

PR #429, opened 2016-11-28
fix(tokens): ttl parameter must be positive

seanmonstar:


vladikoff:
this looks positive to me! (hopefully we won't need to worry about TTL: 0)

PR #430, opened 2016-11-30
fix(deps): update to hapi 16, add shrinkwrap scripts, update other prod deps

vladikoff:
Ref https://github.com/mozilla/fxa-shared/issues/9


vladikoff:
@seanmonstar r?

PR #434, opened 2016-12-19
chore(tests): Uniformly use promises rather than done() callback.

rfk:
The API tests were a strange mixture of returning promises and using the done() callback; this makes them all uniformly use promises.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -193,8 +193,8 @@ describe('/v1', function() {
   describe('/authorization', function() {
 
     describe('GET', function() {
-      it('redirects with all query params', function(done) {
-        Server.api
+      it('redirects with all query params', function() {
+        return Server.api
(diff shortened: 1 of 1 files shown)


rfk:
@mozilla/fxa-devs r?

test/api.js
```diff
@@ -750,7 +757,7 @@ describe('/v1', function() {
           });
         }
 
-        it('should get auth token with secret', function(done){
+        it('should get auth token with secret', function(){
           return getCode(clientId).then(function(code) {
```
shane-tomlinson: I'm a bit surprised there weren't problems with this test earlier since both a promise are returned and `done` specified.

rfk: Yeah me too, the `mocha` docs claim to raise an error if you try to do both.

rfk: Aha, no, what used to get returned here was the result of `.done(done, done)` which is not a promise.

PR #435, opened 2016-12-19
fix(codes): Remove authorization codes after use.

rfk:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1320214; @seanmonstar r?

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -365,7 +365,6 @@ MysqlStore.prototype = {
     });
   },
   getCode: function getCode(code) {
-    logger.debug('getCode');
     var hash = encrypt.hash(code);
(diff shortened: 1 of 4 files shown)


lib/db/mysql/index.js
```diff
@@ -374,8 +373,9 @@ MysqlStore.prototype = {
       return code;
     });
   },
-  removeCode: function removeCode(id) {
-    return this._write(QUERY_CODE_DELETE, [id]);
+  removeCode: function removeCode(code) {
+    var hash = encrypt.hash(code);
+    return this._write(QUERY_CODE_DELETE, [hash]);
```
rfk: AFAICT these would not have worked in their original form, and must hash the code before trying to delete it.

Issue #421, opened 2016-09-28
client-tokens - memory db returns client createdAt instead of token createdAt


vladikoff:
Ref: Memory https://github.com/mozilla/fxa-oauth-server/blob/master/lib/db/memory.js#L256

Ref: SQL https://github.com/mozilla/fxa-oauth-server/blob/5cdf36b4df7c4b0547c96acebf63b9872d71de0f/lib/db/mysql/index.js#L180-L188

We use `MAX(tokens.createdAt)` to sort by newest token in MySQL, but we just pick up `.createdAt` field in the memory db.

PR #436, opened 2016-12-19
fix(memorydb): token createdAt used instead of client createdAt

divyabiyani:
Fixes https://github.com/mozilla/fxa-oauth-server/issues/421

### DIFF
--- a/lib/db/memory.js
+++ b/lib/db/memory.js
@@ -103,16 +103,6 @@ function clone(obj) {
   return clone;
 }
 
-function findByBufProp(arrayOfObjects, prop, value) {
-  for (var i = 0; i < arrayOfObjects.length; i++) {
-    var id = arrayOfObjects[i];
-    if (id[prop].toString('hex') === value) {
-      return id;
-    }
-  }
-  return null;
-}
-
 function deleteByUserId(object, userId) {
   var ids = Object.keys(object);
   for (var i = 0; i < ids.length; i++) {
@@ -260,27 +250,38 @@ MemoryStore.prototype = {
     }
 
     var activeClientIds = [];
-    var activeClients = [];
+    var activeClients = {};
     var ids = Object.keys(this.tokens);
     for (var i = 0; i < ids.length; i++) {
       var id = ids[i];
       if (this.tokens[id].userId.toString('hex') === uid) {
-        activeClientIds.push(this.tokens[id].clientId);
+        activeClientIds.push({
+          clientId: this.tokens[id].clientId,
+          clientTokenTime: this.tokens[id].createdAt
+        });
       }
     }
 
     // unique clients
-    activeClientIds.forEach(function (clientIdBuf) {
-      var clientIdHex = unbuf(clientIdBuf);
-      if (! findByBufProp(activeClients, 'id', clientIdHex)) {
-        var c = self.clients[clientIdHex];
-        if (c.canGrant === false) {
-          activeClients.push(c);
+    activeClientIds.forEach(function (clientTokenObj) {
+      var clientIdHex = unbuf(clientTokenObj.clientId);
+      var client = self.clients[clientIdHex];
+      if (client.canGrant === false) {
+        if (! activeClients[clientIdHex]) {
+          activeClients[clientIdHex] = client;
+        }
+        var clientTokenTime = clientTokenObj.clientTokenTime;
+        if (clientTokenTime > activeClients[clientIdHex].createdAt) {
+          activeClients[clientIdHex].createdAt = clientTokenTime;
         }
       }
     });
 
-    var customSort = activeClients.slice(0);
+    var activeClientsArray = Object.keys(activeClients).map(function (key) {
+        return activeClients[key];
+      });
+
+    var customSort = activeClientsArray.slice(0);
     customSort.sort(function(a, b) {
       if (b.createdAt > a.createdAt) {
         return 1;
--- a/test/api.js
+++ b/test/api.js
@@ -2204,6 +2204,7 @@ describe('/v1', function() {
       });
 
       it('should only list one client for multiple tokens', function() {
+        var tok;
         return getUniqueUserAndToken(client1Id.toString('hex'), {
           uid: user1.uid,
           email: user1.email,
@@ -2223,7 +2224,11 @@ describe('/v1', function() {
               scopes: ['profile']
             });
           })
-          .then(function () {
+          .then(function (client) {
+            return db.getAccessToken(encrypt.hash(client.token));
+          })
+          .then(function (token) {
+            tok = token;
             return Server.api.get({
               url: '/client-tokens',
               headers: {
@@ -2235,6 +2240,7 @@ describe('/v1', function() {
             var result = res.result;
             assert.equal(result.length, 1);
             assert.equal(result[0].id, client1Id.toString('hex'));
+            assert.equal(result[0].lastAccessTime, tok.createdAt.getTime(), 'lastAccessTime should be equal to the latest Token createdAt time');
           });
       });
 


seanmonstar:
This looks like the right direction to me, great! Seems that the tests failed in travis-ci, though.

```
1) /v1 /client-tokens GET /client-tokens should list connected services in set order:
      AssertionError: '54bddae863b4553c' == 'c15f5076200d3555' from assert.equal(result[0].id, client2Id.toString('hex'));
      + expected - actual
      -54bddae863b4553c
      +c15f5076200d3555

2) /v1 /client-tokens GET /client-tokens should only list one client for multiple tokens:
      AssertionError: 4 == 1 from assert.equal(result.length, 1);
      + expected - actual
      -4
      +1
```

lib/db/memory.js
```diff
@@ -265,36 +265,42 @@ MemoryStore.prototype = {
...
+        if (clientTokenTime > activeClients[index].tokenTime) {
+          activeClients[index] = {id: c, tokenTime: clientTokenTime};
         }
       }
     });
 
     var customSort = activeClients.slice(0);
     customSort.sort(function(a, b) {
-      if (b.createdAt > a.createdAt) {
+      if (b.tokenTime > a.tokenTime) {
```
vladikoff: do you think `tokenCreatedAt` is a better name for this than`tokenTime` ?

lib/db/memory.js
```diff
@@ -265,36 +265,42 @@ MemoryStore.prototype = {
     for (var i = 0; i < ids.length; i++) {
       var id = ids[i];
       if (this.tokens[id].userId.toString('hex') === uid) {
-        activeClientIds.push(this.tokens[id].clientId);
+        activeClientIds.push({clientid: this.tokens[id].clientId, clientTokenTime: this.tokens[id].createdAt});
```
vladikoff: once you have an object here I suggest pushing the props to a new line:
```
({
  clientid
})
```

vladikoff: Also make sure to use `clientId`

lib/db/memory.js
```diff
@@ -265,36 +265,42 @@ MemoryStore.prototype = {
...
+      var index = findByBufProp(activeClients, 'id', clientIdHex);
+      var c = self.clients[clientIdHex];
+      if (! Number.isInteger(index)) {
         if (c.canGrant === false) {
-          activeClients.push(c);
+          activeClients.push({id: c, tokenTime: clientTokenTime});
+        }
+      } else {
+        if (clientTokenTime > activeClients[index].tokenTime) {
+          activeClients[index] = {id: c, tokenTime: clientTokenTime};
```
vladikoff: same as https://github.com/mozilla/fxa-oauth-server/pull/436/files#r93253612

vladikoff: Probably not a good idea to use just `c` as a variable for the client. It starts to break down quickly. Such as here: `id: c`.

vladikoff: Ah I'm noticing we already had `c` here, let's fix that up so it is easier to follow in the future 👍

lib/db/memory.js
```diff
@@ -265,36 +265,42 @@ MemoryStore.prototype = {
...
-    activeClientIds.forEach(function (clientIdBuf) {
-      var clientIdHex = unbuf(clientIdBuf);
-      if (! findByBufProp(activeClients, 'id', clientIdHex)) {
-        var c = self.clients[clientIdHex];
+    activeClientIds.forEach(function (clientBuf) {
+      var clientIdHex = unbuf(clientBuf.clientid);
+      var clientTokenTime = unbuf(clientBuf.clientTokenTime);
+      var index = findByBufProp(activeClients, 'id', clientIdHex);
+      var c = self.clients[clientIdHex];
+      if (! Number.isInteger(index)) {
```
vladikoff: I think we need to rethink this part a bit:

```js
+      var index = findByBufProp(activeClients, 'id', clientIdHex);
+      var c = self.clients[clientIdHex];
+      if (! Number.isInteger(index)) {
```

We should probably not use `client.id.name` below. It gives an impression that `id`s have 'names`.

vladikoff:
@divyabiyani as Sean said this is great start and is on the right path to fix the issue! I think the most important part is to fix up the `index` approach, there should be a cleaner away to create the active client list. Also don't forget that you can use `npm test` locally to run the tests.

lib/db/memory.js
```diff
@@ -265,36 +265,42 @@ MemoryStore.prototype = {
...
 
     // unique clients
-    activeClientIds.forEach(function (clientIdBuf) {
-      var clientIdHex = unbuf(clientIdBuf);
-      if (! findByBufProp(activeClients, 'id', clientIdHex)) {
-        var c = self.clients[clientIdHex];
+    activeClientIds.forEach(function (clientBuf) {
+      var clientIdHex = unbuf(clientBuf.clientid);
+      var clientTokenTime = unbuf(clientBuf.clientTokenTime);
+      var index = findByBufProp(activeClients, 'id', clientIdHex);
```
vladikoff: `findByBufProp` means 'find Buffer by property` and then we pass `'id'` into it. Maybe rename it to simplify to something like: `findClientById`. This updated helper function will give you the whole `client` object.

vladikoff: After that you can change `var activeClients = [];` to an object and look up that object by clientId. Instead of using `.push` you would use `activeClients[clientId] = client`

lib/db/memory.js
```diff
@@ -260,27 +250,41 @@ MemoryStore.prototype = {
...
-        var c = self.clients[clientIdHex];
-        if (c.canGrant === false) {
-          activeClients.push(c);
+    activeClientIds.forEach(function (clientBuf) {
+      var clientIdHex = unbuf(clientBuf.clientId);
+      var clientTokenTime = unbuf(clientBuf.clientTokenTime);
+      var client = self.clients[clientIdHex];
+      if (! activeClients[clientIdHex]) {
+        if (client.canGrant === false) {
+          activeClients[clientIdHex] = client;
```
vladikoff: do you need to set the `clientTokenTime` here too?

lib/db/memory.js
```diff
@@ -260,27 +250,41 @@ MemoryStore.prototype = {
...
+          activeClients[clientIdHex].createdAt = clientTokenTime;
         }
       }
     });
 
-    var customSort = activeClients.slice(0);
+    var activeClientsArray = Object.keys(activeClients).map(
+      function (key) {
+        return activeClients[key];
+      });
```
vladikoff: style nit: 

```js
var activeClientsArray = Object.keys(activeClients).map(function (key) {
  return activeClients[key];
});
```

vladikoff:
Fixes https://github.com/mozilla/fxa-oauth-server/issues/421

lib/db/memory.js
```diff
@@ -260,27 +250,39 @@ MemoryStore.prototype = {
...
+        }
+        var clientTokenTime = unbuf(clientBuf.clientTokenTime);
+        if (clientTokenTime > activeClients[clientIdHex].createdAt) {
+          activeClients[clientIdHex].createdAt = clientTokenTime;
         }
       }
     });
 
-    var customSort = activeClients.slice(0);
+    var activeClientsArray =
```
vladikoff:
```
+    var activeClientsArray = Object.keys(activeClients).map(function (key) {
```

This can just be a single line, no need to line break after `=`

vladikoff:
@seanmonstar any other feedback from you?

manually testing with memory db `createdAt` now shows latest token instead of client created date

lib/db/memory.js
```diff
@@ -260,27 +250,39 @@ MemoryStore.prototype = {
...
-        if (c.canGrant === false) {
-          activeClients.push(c);
+    activeClientIds.forEach(function (clientBuf) {
+      var clientIdHex = unbuf(clientBuf.clientId);
+      var client = self.clients[clientIdHex];
+      if (client.canGrant === false) {
+        if (! activeClients[clientIdHex]) {
+          activeClients[clientIdHex] = client;
+        }
+        var clientTokenTime = unbuf(clientBuf.clientTokenTime);
```
vladikoff: @divyabiyani do you need to `unbuf` here? seems like this is not a `Buffer`

lib/db/memory.js
```diff
@@ -260,27 +250,39 @@ MemoryStore.prototype = {
...
     }
 
     // unique clients
-    activeClientIds.forEach(function (clientIdBuf) {
-      var clientIdHex = unbuf(clientIdBuf);
-      if (! findByBufProp(activeClients, 'id', clientIdHex)) {
-        var c = self.clients[clientIdHex];
-        if (c.canGrant === false) {
-          activeClients.push(c);
+    activeClientIds.forEach(function (clientBuf) {
```
seanmonstar: The name `clientIdBuf` was to help point out that it was a `Buffer` of the client ID. Seems that it no longer is, so it could probably be renamed to just `client` or similar.

test/api.js
```diff
@@ -2235,6 +2240,7 @@ describe('/v1', function() {
             var result = res.result;
             assert.equal(result.length, 1);
             assert.equal(result[0].id, client1Id.toString('hex'));
+            assert.equal(result[0].lastAccessTime, tok.createdAt.getTime());
```
vladikoff: move the `lastAccessTime should equal to the latest Token createdAt time`from the test title into here: 

`assert.equal(result[0].lastAccessTime, tok.createdAt.getTime(), 'lastAccessTime should equal to the latest Token createdAt time');

vladikoff:
awesome work @divyabiyani !! 🍰

PR #439, opened 2017-01-13
fix(scripts): Use pure JS module to generate RSA keypairs

jbuck:
This PR switches from a native module to a pure JS module. No build dependencies necessary!


vladikoff:
Let's do a quick check @seanmonstar or @jrgm if this modules is good to use, I think we had some issues in the past when dealing with this in the auth-server repo.

vladikoff:
It seems like this work! I tested the generated keys and the server. 🔖

PR #440, opened 2017-01-14
fix(headers): add cache-control headers to api endpoints; extend tests

jrgm:
r? - @seanmonstar - /cc @g-k 

This is akin to the patch to profile server to set an extended cache-control header (and extend the tests to check for all the security headers).

### DIFF
--- a/test/lib/util.js
+++ b/test/lib/util.js
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const assert = require('insist');
+
+function assertSecurityHeaders(res) {
+  var expect = {
+    'strict-transport-security': 'max-age=15552000; includeSubDomains',
+    'x-content-type-options': 'nosniff',
+    'x-xss-protection': '1; mode=block',
+    'x-frame-options': 'DENY',
+    'cache-control': 'private, no-cache, no-store, must-revalidate'
+  };
+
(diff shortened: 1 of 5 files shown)


test/lib/util.js
```diff
@@ -0,0 +1,24 @@
...
+
+const assert = require('insist');
+
+function assertSecurityHeaders(res) {
+  var expect = {
+    'strict-transport-security': 'max-age=15552000; includeSubDomains',
+    'x-content-type-options': 'nosniff',
+    'x-xss-protection': '1; mode=block',
+    'x-frame-options': 'DENY',
+    'cache-control': 'private, no-cache, no-store, must-revalidate'
```
shane-tomlinson: Does it make sense to hook the cache-control value here up to config so that if config changes, the test doesn't need to be modified?

vladikoff: Yea, i think we do this in other servers @jrgm

jrgm: Updated so "cache-control" is a config option.

test/lib/util.js
```diff
@@ -0,0 +1,25 @@
...
+    'x-content-type-options': 'nosniff',
+    'x-xss-protection': '1; mode=block',
+    'x-frame-options': 'DENY'
+  };
+
+  Object.keys(expect).forEach(function(header) {
+    assert.equal(res.headers[header], expect[header]);
+  });
+
+  assert.equal(res.headers['cache-control'], config.cacheControl);
```
shane-tomlinson: 👍

shane-tomlinson:
I think it's beautiful, but someone with more OAuth server experience should double check. Thanks @jrgm!

jrgm:
r? - @seanmonstar

PR #441, opened 2017-01-16
refactor(client): scope added in memory

### LINKED ISSUE #431: return granted scopes used in client-tokens list
We should show the user the scopes that were used when they logged in into some relier.
We should also show those somehow in the Devices and Apps list.

divyabiyani:
Fixes #431.

### DIFF
--- a/lib/db/memory.js
+++ b/lib/db/memory.js
@@ -7,6 +7,7 @@ const unbuf = require('buf').unbuf.hex;
 
 const config = require('../config');
 const encrypt = require('../encrypt');
+const helpers = require('./helpers');
 const logger = require('../logging')('db.memory');
 const P = require('../promise');
 const unique = require('../unique');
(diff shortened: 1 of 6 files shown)


lib/db/memory.js
```diff
@@ -269,7 +270,14 @@ MemoryStore.prototype = {
       if (client.canGrant === false) {
         if (! activeClients[clientIdHex]) {
           activeClients[clientIdHex] = client;
+          activeClients[clientIdHex].scope = [];
         }
+        clientTokenObj.clientScope.forEach(function (scope) {
```
vladikoff: @divyabiyani This needs to handle this transformation:

with 3 token scopes of 1) `clients:write` 2) `profile profile:write` 3) `profile` you need to end up with: 

`clients:write profile profile:write`. 

Would this code here become:  `clients:write profile profile:write profile` ?

divyabiyani: No, @vladikoff. It is giving scope="clients:write,profile,profile:write"

vladikoff: Oh sorry I got confused with the way scopes are stored. Let's add a test for this new behaviour. 
We need the test to issue ~3-4 tokens with different scopes and return the unified list.

lib/db/mysql/index.js
```diff
@@ -431,7 +430,67 @@ MysqlStore.prototype = {
...
+          }
+        });
+
+        var clientTokenTime = clientTokenObj.createdAt;
+        if (clientTokenTime > activeClients[clientIdHex].createdAt) {
+          activeClients[clientIdHex].createdAt = clientTokenTime;
+        }
+      });
+      var activeClientsArray = Object.keys(activeClients).map(function (key) {
+        var Scopes = activeClients[key].scope;
```
vladikoff: no need to make `Scopes`, no need for the capital `S` in there...

lib/db/mysql/index.js
```diff
@@ -431,7 +430,67 @@ MysqlStore.prototype = {
   getActiveClientTokensByUid: function getActiveClientTokensByUid(uid) {
     return this._read(QUERY_ACTIVE_CLIENT_TOKENS_BY_UID, [
       buf(uid)
-    ]);
+    ]).then(function(activeClientIds) {
+      var activeClients = {};
```
vladikoff: do you think we can make a single function that will work for both (memory and MySQL engines)?

test/api.js
```diff
@@ -2286,13 +2286,13 @@ describe('/v1', function() {
         return getUniqueUserAndToken(client1Id.toString('hex'), {
           uid: user1.uid,
           email: user1.email,
-          scopes: ['profile']
+          scopes: ['profile', 'profile:write']
```
vladikoff: @divyabiyani could we please add a new test just for this with more advanced cases, such as:

Same client:

1) scopes: ['profile', 'profile:write']
2) scopes: ['clients:write']
3) scopes: ['basket:write', 'profile:email']
4) scopes: ['profile:uid', 'profile', 'profile:write']

In this new test we just care about scopes, so we don't need the extra assertions.

divyabiyani: Ok @vladikoff . I will add it.

lib/db/mysql/index.js
```diff
@@ -12,6 +12,7 @@ const MysqlPatcher = require('mysql-patcher');
 const config = require('../../config');
 const encrypt = require('../../encrypt');
 const P = require('../../promise');
+const scope = require('../scope');
 const Scope = require('../../scope');
```
vladikoff: @divyabiyani hm the `scope` and `Scope` would be really confusing, we need to rename of them

lib/db/scope.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const unbuf = require('buf').unbuf.hex;
+
+const Scope = {
```
vladikoff: @divyabiyani does it make sense to call this `helpers` or something? and rename this to `lib/db/helpers.js`?

We also need a unit test for every function in this module

vladikoff: you can probably just do `module.exports = {
// helpers functions and code here
}

lib/db/scope.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const unbuf = require('buf').unbuf.hex;
```
vladikoff: let's add a comment above here to explain what this file is for. 
Example: `This module consists of helpers that are used by both memory and MySQL database engines`

vladikoff:
Continued in https://github.com/mozilla/fxa-oauth-server/pull/445

PR #442, opened 2017-01-23
fix(keys): Generate unique 'kid' field when regenerating JWK keys

### LINKED ISSUE #393: Use fresh "kid" when generating fresh keys
Our gen_keys.js script currently uses a static `kid: "2015.12.16-1"` for all keys that it generates.  In the interests of good key-management hygiene, we should generate a fresh key id for each new key.  Options include using a hash of the key data, or a timestamp.

No big deal, just something I noticed.

rfk:
Fixes #393; @seanmonstar r?

### DIFF
--- a/scripts/gen_keys.js
+++ b/scripts/gen_keys.js
@@ -21,6 +21,7 @@
 
 const fs = require('fs');
 const assert = require('assert');
+const crypto = require('crypto');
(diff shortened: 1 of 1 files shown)


seanmonstar:
Woops, I restarted the Travis build this morning and then forgot to check back. Looks good!

PR #448, opened 2017-02-10
fix(db): ensure strict mode

### LINKED ISSUE #446: Ensure db connections are always in strict mode
Over in the profile-server, we recently added some logic to force all db connections into strict mode:

  https://github.com/mozilla/fxa-profile-server/pull/221

It seems to be working well in production, so let's apply it to oauth-server as well.  It might make sense to factor it out into a common helper lib, I haven't looked at the details of how much code we could share.

vladikoff:
Fixes #446

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -18,15 +18,21 @@ const unique = require('../../unique');
 const patch = require('./patch');
 
 const MAX_TTL = config.get('expiration.accessToken');
+const REQUIRED_SQL_MODES = [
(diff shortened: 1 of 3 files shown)


vladikoff:
@seanmonstar this one seemed fine: https://github.com/mozilla/fxa-profile-server/pull/221 

Let's get a final r from @rfk

rfk:
> will the change to the charset have no problems?

My understanding is that it should be fine, since it's only setting the default for the connection, not trying to change any existing db metadata.

PR #454, opened 2017-04-03
fix(test): fix unhandled rejection error with memory db impl

jrgm:
I wasn't returning the promise[1]. Also, the memory db impl should be expected to pass, not fail.

[1] e.g., https://travis-ci.org/mozilla/fxa-oauth-server/jobs/213525486#L677

r? - @seanmonstar

### DIFF
--- a/test/db/index.js
+++ b/test/db/index.js
@@ -78,7 +78,7 @@ describe('db', function() {
 
     it('2-byte encoding preserved', makeTest(randomString(8), 'Düsseldorf'));
     it('3-byte encoding preserved', makeTest(randomString(8), '北京')); // Beijing
-    it('4-byte encoding throws', function() {
+    it('4-byte encoding throws with mysql; ok with memdb', function() {
       var data = {
         id: randomString(8),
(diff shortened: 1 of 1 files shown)


test/db/index.js
```diff
@@ -90,14 +90,22 @@ describe('db', function() {
         trusted: true
       };
 
-      db.registerClient(data)
+      return db.registerClient(data)
         .then(function(c) {
-          assert.fail('This should not have succeeded.');
+          if (config.get('db.driver') === 'memory') {
+            assert.ok(true, 'Everything is cool. Everythings ok.');
```
jrgm: I should probably update this PR to assert that the returned value is equal to the data value if using memory db, not just that it didn't throw.

jrgm: updated to check that 4-byte with memdb works ok; but mysql throws.

jrgm:
updated to check that 4-byte with memdb works ok; but mysql throws.

jrgm:
The travis tests (or is it also circle-ci) keep getting rando failures like `#_initialClients
fxa-oauth-server.db.mysql.ERROR: updateDbSchema Error: ER_DUP_FIELDNAME: Duplicate column name 'authAt'`. 

Nothing to do with this change, ISTM.

PR #459, opened 2017-04-14
fix(patcher): Fix patcher with no pre-loaded clients

jbuck:


jrgm:
Just to note, if clients is `[]` (the default), no promise if returned and undefined.then() crashes subsequently. This just ensures there is always a promise returned.

Issue #465, opened 2017-06-03
Implement PKCE to support public clients using the code grant flow


rfk:
@vladikoff I'm filing this to capture the task from our meeting with @Natim yesterday.  Per that discussion:

* The hoverpad webextension and mobile app must be considered "public clients".  This is OAuth spec terminology for "clients which can't keep anything confidential, and so can't meaningfully be authenticated with a client_secret".
* We want to support such clients using the code-based login flow, so that they can obtain encryption keys.
* Let's hew to existing standards and best practices for allowing such clients in a secure way.

In practice I think that means implementing the "Proof Key for Code Exchange" spec [1].  It's specified in excruciating detail in the RFC but the core idea is that:

* When initiating the OAuth dance, the client generates and remembers a random secret called the `code_verifier`. They provide `code_challenge = HASH(code_verifier)` as an additional input parameter to the flow, and we remember this value for later use.
* At the end of the dance, when the client has received the `code` parameter is wants to exchange it for a token, they must provide the original value of `code_verifier` along side it.
* On the server, we receive `code` and `code_verifier`, and check that `HASH(code_verifier)` matches the value of `code_challenge` that was submitted at the beginning of the flow.

The end result is that only the actual client that initiated the login flow is able to redeem the authorization code at the end, so we can protect public clients against other apps snooping their login flow.  For example, this will prevent other webextensions from snooping and using the returned authorization code.

IIUC the required changes from the oauth-server perspective will include:

* Adjust the db schema for client records, to be able to mark some of them as "public clients" that do not have a client_secret.
* At the start of the flow in `GET /authorization`, accept and remember the new `code_challenge` parameter.
* In the middle of the flow, in `POST /authorization` where we generated an authorization code, ensure that the code gets associated with the correct value of `code_challenge`. 
* At the end of the flow in `POST /token`:
    * accept the new `code_verifier` parameter
    * check that it matches the `code_challenge` associated with the incoming authorization code

[1] https://tools.ietf.org/html/rfc7636

vladikoff:
This is a good article for this: https://www.authlete.com/documents/article/pkce/index

seanmonstar:
> Adjust the db schema for client records, to be able to mark some of them as "public clients" that do not 
have a client_secret.

I believe this is in the DB already, though perhaps with a different name. The `canGrant` column was meant for clients that don't have a secret, but do the implicit auth grant flow, since they can mint a browserid assertion, which we trust instead of a secret.

[...]

PR #466, opened 2017-06-08
feat(pkce): add PKCE support to the oauth server

vladikoff:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/465

TODO:
- [x] MySQL Support 
- [x] Tests
- [x] Docs


config/dev.json
```diff
@@ -15,6 +15,16 @@
       "canGrant": false
     },
     {
+      "id": "38a6b9b3a65a1871",
+      "hashedSecret": "289a885946ee316844d9ffd0d725ee714901548a1e6507f1a40fb3c2ae0c99f1",
```
vladikoff: @rfk to avoid making the secret `optional` in several places, instead we can still assign a secret to these clients but prevent them from using it if they are `publicClient: true`

Thoughts?

lib/config.js
```diff
@@ -313,6 +313,10 @@ const conf = convict({
       doc: 'Bytes of generated client_secrets',
       default: 32
     },
+    codeVerifier: {
+      doc: 'Bytes of generated PKCE code_verifier',
+      default: 32
```
vladikoff: TODO: this value needs changing

lib/routes/authorization.js
```diff
@@ -65,7 +65,11 @@ function generateCode(claims, client, scope, req) {
     email: claims['fxa-verifiedEmail'],
     scope: scope,
     authAt: claims['fxa-lastAuthAt'],
-    offline: req.payload.access_type === ACCESS_TYPE_OFFLINE
+    offline: req.payload.access_type === ACCESS_TYPE_OFFLINE,
+    // PKCE options
+    code_challenge_method: req.payload.code_challenge_method,
+    code_challenge: req.payload.code_challenge,
+
```
vladikoff: TODO: clean up

lib/routes/authorization.js
```diff
@@ -142,6 +146,9 @@ module.exports = {
         .valid(ACCESS_TYPE_OFFLINE, ACCESS_TYPE_ONLINE)
         .default(ACCESS_TYPE_ONLINE)
         .optional(),
+      // PKCE support
+      code_challenge_method: Joi.string().optional(), // Force 'S256'
```
vladikoff: @rfk thoughts on only supporting the `S256` and not `plain` - code challenge method?

lib/routes/token.js
```diff
@@ -127,7 +138,10 @@ module.exports = {
     // we don't use, such as `response_type`, or something else. Instead
     // of giving an error here, we can just ignore them.
     payload: function validatePayload(value, options, next) {
-      return Joi.validate(value, PAYLOAD_SCHEMA, { stripUnknown: true }, next);
+      return Joi.validate(value, PAYLOAD_SCHEMA, { stripUnknown: true }, function(err, value) {
+        console.log(err);
+        next(err, value)
```
vladikoff: I'm surprised we don't have this right now, the detailed validation fails just kind of get dropped and are not logged it seems.

vladikoff: need to add a better logging statement to that though

lib/routes/token.js
```diff
@@ -138,17 +152,24 @@ module.exports = {
...
+          return confirmPkceClient(params.client_id)
+            .then(function() {
+              return confirmPkceCode(params.client_id, params.code, params.code_verifier);
+            });
+        } else {
+          return confirmClient(params.client_id, params.client_secret)
+            .then(function() {
+              return confirmCode(params.client_id, params.code);
+            });
+        }
```
vladikoff: TODO: above can be cleaned up

lib/routes/token.js
```diff
@@ -174,6 +195,49 @@ module.exports = {
   }
 };
 
+function base64URLEncode(str) {
```
vladikoff: TODO: move this to some util module

lib/routes/token.js
```diff
@@ -174,6 +195,49 @@ module.exports = {
...
+  return str.toString('base64')
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=/g, '');
+}
+
+function pkceHash(input) {
```
vladikoff: TODO: move this to some util module

lib/routes/token.js
```diff
@@ -174,6 +195,49 @@ module.exports = {
...
+      logger.debug('code.notFound', { code: code });
+      throw AppError.unknownCode(code);
+    }
+
+    if (pkceHash(pkceVerifier) === codeObj.code_challenge) {
+      return db.removeCode(buf(code)).then(function() {
+        return codeObj;
+      });
+    } else {
+      throw AppError.unknownCode(code);
```
vladikoff: TODO: better error here

lib/routes/token.js
```diff
@@ -174,6 +195,49 @@ module.exports = {
...
+function confirmPkceClient(id) {
+  return db.getClient(buf(id)).then(function(client) {
+    if (!client) {
+      logger.debug('client.notFound', { id: id });
+      throw AppError.unknownClient(id);
+    }
+
+    if (!client.publicClient) {
+      logger.debug('client.notPublicClient', { id: id });
+      throw AppError.unknownClient(id); // TODO: custom error?
```
vladikoff: TODO: need to add a new error type here

lib/validators.js
```diff
@@ -16,7 +16,9 @@ exports.clientId = Joi.string()
 exports.clientSecret = Joi.string()
   .length(config.get('unique.clientSecret') * 2) // hex = bytes*2
   .regex(exports.HEX_STRING)
-  .required();
```
vladikoff: this needs to come back, but removed because the validation rules are complex, needs to be fixed

lib/routes/authorization.js
rfk: Unless the spec mandates that we support `plain`, I am +100

vladikoff: Spec says:

> If the client is capable of using "S256", it MUST use "S256", as
   "S256" is Mandatory To Implement (MTI) on the server.  Clients are
   permitted to use "plain" only if they cannot support "S256" for some
   technical reason and know via out-of-band configuration that the
   server supports "plain".

rfk: Then I am +100 to only support `S256` :-)

config/dev.json
rfk: Seems OK to me; looking in the file below, seems like we also took this approach with the `canGrant: true` clients which would also be expected not to hold a secret.

lib/config.js
rfk: I'm not sure it's useful having this in config, as we don't control the generation of this token.  The spec says we need to support "a minimum length of 43 characters and a maximum length of 128 characters."

lib/routes/authorization.js
rfk: Note that the spec says this defaults to `plain` if unspecified, so I think we'll need to insist on it being present with a value of `S256`.

lib/routes/token.js
```diff
@@ -53,6 +54,19 @@ const PAYLOAD_SCHEMA = Joi.object({
...
+    .when('grant_type', {
+      is: GRANT_JWT,
+      then: Joi.forbidden(),
+      //otherwise: Joi.required()
+    }),
+    // .when('code_verifier', {
+    //   is: Joi.string(),
+    //   then: Joi.forbidden(),
+    //   //otherwise: Joi.required()
+    // }),
```
vladikoff: @seanmonstar any ideas here how to make `clientSecret` required in ALL cases except when `GRANT_JWT` or `code_verifier` is SET? The commented code above still wants `client_secret`

lib/db/mysql/patches/patch-017-018.sql
```diff
@@ -0,0 +1,11 @@
+--  Add `publicClient` column to the `clients` table.
+
+ALTER TABLE clients ADD COLUMN publicClient BOOLEAN DEFAULT FALSE NOT NULL AFTER canGrant;
+UPDATE clients SET publicClient=false;
+--  Add `authAt` column to the `codes` table.
```
rfk: nit: looks like this comment copypata'd the mention of the `authAt` column.

lib/db/mysql/patches/patch-017-018.sql
```diff
@@ -0,0 +1,11 @@
...
+
+ALTER TABLE codes
+ADD COLUMN code_challenge_method VARCHAR(256) AFTER offline,
+ADD COLUMN code_challenge VARCHAR(256) AFTER code_challenge_method;
```
rfk: Should we use `ALGORITHM = INPLACE, LOCK = NONE;` here for safety?

vladikoff:
@seanmonstar @rfk please r?

lib/db/mysql/patches/patch-017-018.sql
```diff
@@ -0,0 +1,11 @@
+--  Add `publicClient` column to the `clients` table.
+ALTER TABLE clients ADD COLUMN publicClient BOOLEAN DEFAULT FALSE NOT NULL AFTER canGrant;
+UPDATE clients SET publicClient=false;
+
+--  Add `code_challenge_method` and `code_challenge` column to the `codes` table.
```
seanmonstar: These new column names here are snake case, all other column names are camel case.

lib/error.js
```diff
@@ -232,4 +232,36 @@ AppError.expiredToken = function expiredToken(expiredAt) {
...
+    code: 400,
+    error: 'Bad Request',
+    errno: 117,
+    message: 'Incorrect code_challenge'
+  }, {
+    requestCodeChallenge: code_challenge
+  });
+};
+
+AppError.missingPkceParameters = function invalidScopes() {
```
seanmonstar: Function name of `invalidScopes` doesn't match the error. Likely a left over from copying a previous error constructor.

test/routes/authorization.js
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  if (! fail) {
+    throw new Error('Did not throw!');
+  }
+}
+
+describe('/authorization POST', function () {
+  it('fails with no client_id', () => {
+    joiAssertFail({
```
seanmonstar: YES!

PR #467, opened 2017-06-12
fix(tests): speed up and upgrade the test runner

vladikoff:
@seanmonstar r?

* The grunt task test runner is a little slow and makes running specifc tests hard.
* Our coverage tool was out of date and has problems with ES6. 
* Mocha also updated to latest.

We also have an issue where the server (for the api.js test) takes a bit of time to start up. This PR is a first step to start fixing that.


lib/routes/root.js
```diff
@@ -35,9 +35,8 @@ module.exports = {
     var gitDir = path.resolve(__dirname, '..', '..', '.git');
     exec('git rev-parse HEAD', { cwd: gitDir }, function(err, stdout) {
       commitHash = stdout.replace(/\s+/, '');
-      var configPath = path.join(gitDir, 'config');
       var cmd = 'git config --get remote.origin.url';
-      exec(cmd, { env: { GIT_CONFIG: configPath } }, function(err, stdout) {
+      exec(cmd, { cwd: gitDir }, function(err, stdout) {
```
seanmonstar: The auth server uses the env variables. I feel that was changed specifically because of some security concern...

vladikoff:
> What if we had a new task that just called the test-local script instead?

I don't like putting the test stuff into grunt tasks because of the nature of tests. 

* running specific files
* different envs
* harder to attach the debugger
* another layer of complexity

lib/routes/root.js
vladikoff: When I run `npm test` this command says `git is not a command` without this change. It's strange

vladikoff: The other `git` command works above with the set cwd

vladikoff:
> So, specifically the usage of mochaTest is slow?

It is a bit slow, this PR removes the "loading of tasks" slowness and the old coverage tool which helps a bit. We can speed up the server loading in the future PRs

lib/routes/root.js
seanmonstar: Relevant PR: https://github.com/mozilla/fxa-auth-server/pull/1715

Looks like this is used to fix the tests: `exec(cmd, { env: { GIT_CONFIG: configPath, PATH: process.env.PATH } }`

vladikoff: Perfect!

seanmonstar:
> I don't like putting the test stuff into grunt tasks because of the nature of tests.

I wanted to take advantage of `time-grunt`, and better deal with the list of config values living in our package.json file. I'm fine this change in the oauth server here, I'll only submit a PR for the auth-server if I find a solution for the problems you listed.

> "loading of tasks" slowness

That still exists because of calling `grunt eslint copyright` after, right?

> speed up the server loading in the future PRs

I haven't measure in this repo if there is an operation that could be cached. I did notice while in the auth server that simply calling `server.route(routes)` is super slow, as it compiles all the validation and sets up a route matcher for parameters...

vladikoff:
> I did notice while in the auth server that simply calling server.route(routes) is super slow

Exactly, I think we can find some things to optimize in this server. Would this be okay to merge once I fix the https://github.com/mozilla/fxa-oauth-server/pull/467/files/848986c856d7d062fda1ea5ab77eb579b4449d49#diff-6c72c7fb241ef4b39f1a2968a2411315 ?

seanmonstar:
Yep!

vladikoff:
> Yep!

Thanks @seanmonstar !

PR #476, opened 2017-08-14
feat(pkce): add ability for PKCE clients to use refresh_tokens

### LINKED ISSUE #472: Allow token refresh using the PKCE flow
Need to handle the case here: https://github.com/mozilla/fxa-oauth-server/blob/52d935183b7bbd5813f52def13ac14c22acd162a/lib/routes/token.js#L177 where clients don't have a client_secret and want to renew their poor token 💻

vladikoff:
Fixes #472

@rfk @seanmonstar r?

### DIFF
--- a/lib/routes/token.js
+++ b/lib/routes/token.js
@@ -62,6 +62,10 @@ const PAYLOAD_SCHEMA = Joi.object({
     .when('code_verifier', {
       is: Joi.string().required(), // if (typeof code_verifier === 'string') {
       then: Joi.forbidden()
+    })
+    .when('grant_type', {
+      is: GRANT_REFRESH_TOKEN,
+      then: Joi.optional()
     }),
 
   code_verifier: validators.codeVerifier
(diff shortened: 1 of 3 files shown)


lib/routes/token.js
```diff
@@ -175,9 +179,31 @@ module.exports = {
...
+          var confirmClientPromise;
+
+          if (dbClient.publicClient) {
+            if (params.client_secret) {
+              throw new AppError.invalidRequestParameter('client_secret');
+            }
+
+            confirmClientPromise = confirmPkceClient(params.client_id);
+          } else {
+            confirmClientPromise = confirmClient(params.client_id, params.client_secret);
```
seanmonstar: The `confirmClient` does another db lookup, but this function already did one, and then compares a hash of the secrets.

vladikoff:
> It seems to me that the confirmBlahClient stuff could be refactored ever so slightly. So as to not have secret comparison code duplicated, that could just be 1 function, used both here and in the original confirmClient function.

Thanks, that's a good plan!

vladikoff:
@seanmonstar How about now? I tried moved the db code out of the existing functions. Let me know..

vladikoff:
Example usage: 

POST http://127.0.0.1:9010/v1/token

```
{
  "refresh_token": "8425d9248b92f86d7bf518ee6dc51e304c8702892bbe2466e487459c05d64628",
  "client_id": "c6d74070a481bc10",
  "grant_type": "refresh_token",
  "scope": "profile%20https%3A%2F%2Fidentity.mozilla.org%2Fapps%2Fnotes"
}
```

Issue #481, opened 2017-09-08
Revoke tokens for `canGrant` and `isPublic` clients when users changes or resets their password


rfk:
As part of integrating Firefox Notes with scoped keys, we'll need to invalidate its OAuth tokens whenever the user changes or resets their password.

@vladikoff and I discussed this in our 1:1 the other day, and while we can iterate on the exact policy as we go, a good starting point seems to be: "when the user changes or resets their password, all tokens for `canGrant` and `isPublic` clients are revoked".

This will have the happy effect of also clearing oauth tokens issued by content-server for its own use, fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1374032 at the same time.

The oauth-server already receives notification of password resets via SQS, but I think we'll have to add a new event type for password change.  Ref https://github.com/mozilla/fxa-auth-server/issues/2103

PR #485, opened 2017-09-18
feat(tokens): add support for password change and reset event

vladikoff:
Fixes https://github.com/mozilla/fxa-oauth-server/issues/481
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1374032

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -356,9 +356,6 @@ MysqlStore.prototype = {
   getClients: function getClients(email) {
     return this._read(QUERY_CLIENT_LIST, [ email ]);
   },
-  getPublicClients: function getPublicClients() {
-    return this._read(QUERY_PUBLIC_CLIENTS_LIST);
-  },
   removeClient: function removeClient(id) {
     return this._write(QUERY_CLIENT_DELETE, [buf(id)]);
   },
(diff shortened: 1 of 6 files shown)


vladikoff:
@rfk @vbudhram r?

lib/db/mysql/index.js
```diff
@@ -606,6 +610,23 @@ MysqlStore.prototype = {
...
+   * @param userId
+   * @returns {Promise}
+   */
+  removePublicAndCanGrantTokens: function removePublicAndCanGrantTokens(userId) {
+    const uid = buf(userId);
+
+    return this._read(QUERY_PUBLIC_CLIENTS_LIST).then((_clients) => {
+      const clientIds = _clients.map((client) => client.id);
+
+      return this._write(QUERY_DELETE_ACCESS_TOKEN_FOR_PUBLIC_CLIENTS, [uid, clientIds])
```
rfk: Looking up the client ids in a separate query is fine, but you could skip a round-trip by using MySQL's neat ability to delete through a join:

  https://stackoverflow.com/questions/652770/delete-with-join-in-mysql#4192849

It'd look something like (untested):

```
DELETE tokens
FROM tokens INNER JOIN clients ON tokens.clientId = clients.id
WHERE tokens.userId = ? AND (clients.publicClient = 1 OR clients.canGrant = 1)
```

test/db/index.js
```diff
@@ -397,6 +397,100 @@ describe('db', function() {
...
+    it('revokes tokens for publicClient', () => {
+      return testRemovalWithClient({
+        publicClient: true
+      }).then((tokens) => {
+        assert.equal(tokens[0], undefined);
+        assert.equal(tokens[1], undefined);
+      });
+    });
+
+    it('does not revokesTokens for not canGrant or not publicClient', () => {
```
rfk: nit: s/or/and/ ?

lib/db/mysql/index.js
vladikoff: > MySQL's neat ability

Let's stay away from that 🤕  IMHO

PR #486, opened 2017-09-20
feat(keys): add keys_jwe support

### LINKED ISSUE #484: Add ability to return `keys_jwe` field in authorization response
In support of [1], the oauth-server needs the ability to accept an encrypted bundle of key material from content-server during the oauth dance, and provide it to the relier at completion.  We'll need to:

* Accept a "keys_jwe" input parameter when provisioning an authorization code, and store it in the db alongside other metadata for the code
* When the relier exchanges the authorization code for a token, return the saved "keys_jwe" bundle alongside the other values in the response

[1] https://docs.google.com/document/d/1IvQJFEBFz0PnL4uVlIvt8fBS_IPwSK-avK0BRIHucxQ/edit#heading=h.mqdsr2ucjm61

vladikoff:
Fixes #484

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -321,7 +321,7 @@ back to the client. This code will be traded for a token at the
 - `access_type`: Optional. A value of `offline` will generate a refresh token along with the access token.
 - `code_challenge_method`: Required if using [PKCE](pkce.md). Must be `S256`, no other value is accepted.
 - `code_challenge`: Required if using [PKCE](pkce.md). A minimum length of 43 characters and a maximum length of 128 characters string, encoded as `BASE64URL`.
-- `keys_jwe`: A JWE bundle of the request.
+- `keys_jwe`: Optional. A JWE bundle to be returned to the client when it redeems the authorization code.
 
 
 **Example:**
@@ -421,7 +421,7 @@ A valid request will return a JSON response with these properties:
 - `token_type`: A string representing the token type. Currently will always be "bearer".
 - `auth_at`: An integer giving the time at which the user authenticated to the Firefox Accounts server when generating this token, as a UTC unix timestamp (i.e.  **seconds since epoch**).
 - `id_token`: (Optional) If the authorization was requested with `openid` scope, then this property will contain the OpenID Connect ID Token.
-- `keys_jwe`: (Optional) A JWE bundle from the authorization request.
+- `keys_jwe`: (Optional) Returns the JWE bundle that if the authorization request had one.
 
 **Example:**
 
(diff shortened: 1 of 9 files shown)


vladikoff:
@rfk r?

docs/api.md
```diff
@@ -321,6 +321,7 @@ back to the client. This code will be traded for a token at the
 - `access_type`: Optional. A value of `offline` will generate a refresh token along with the access token.
 - `code_challenge_method`: Required if using [PKCE](pkce.md). Must be `S256`, no other value is accepted.
 - `code_challenge`: Required if using [PKCE](pkce.md). A minimum length of 43 characters and a maximum length of 128 characters string, encoded as `BASE64URL`.
+- `keys_jwe`: A JWE bundle of the request.
```
rfk: This could use a few more words; suggestion: "(Optional) A JWE bundle to be returned to the client when it redeems the authorization code`.

lib/db/mysql/patches/patch-018-019.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE codes ADD COLUMN keysJwe VARCHAR(1024),
```
rfk: Are we confident `1024` will be enough for anybody, or should be make this a `MEDIUMTEXT` or similar column?  Either way, let's ensure we set a length limit on it during request validation so it doesn't get truncated in the db.

lib/routes/authorization.js
```diff
@@ -160,6 +161,12 @@ module.exports = {
           is: CODE,
           then: Joi.optional(),
           otherwise: Joi.forbidden()
+        }),
+      keys_jwe: Joi.string()
```
rfk: Add length restriction here?

rfk: We could also validate that it's only b64 encoded characters, or whatever format restrictions we can validate cheaply.

test/api.js
```diff
@@ -780,6 +780,53 @@ describe('/v1', function() {
...
+            url: '/token',
+            payload: {
+              client_id: client2.id.toString('hex'),
+              code: code,
+              code_verifier: code_verifier
+            }
+          });
+        }).then((res) => {
+          assert.equal(res.statusCode, 200);
+          assert.equal(res.result.keys_jwe, keys_jwe);
```
rfk: Please also add a test that the normal non-keys flow does *not* return a `keys_jwe` member, for completeness.

vladikoff:
@rfk updated!

lib/validators.js
```diff
@@ -35,3 +35,6 @@ exports.assertion = Joi.string()
   .max(10240)
   .regex(/^[a-zA-Z0-9_\-\.~=]+$/);
 
+exports.jwe = Joi.string()
+  .max(1024)
+  .regex(/^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/);
```
vladikoff: add a comment for the regex

test/api.js
vladikoff: from mtg: new test cases seem fine

PR #487, opened 2017-09-21
feat(scopes): add key-data and scope support

### LINKED ISSUE #483: Add API for returning scoped key metadata
In support of [1], the oauth-server needs an API that can accept as input:

* Client metadata
* The set of scopes being requested
* A user's identity assertion 

And return the necessary metadata for deriving scoped keys:

* Canoncalized set of scopes that have associated keys
* For each scope, the salt to use when deriving the keys
* For each scope, the timestamp at which the key last changed

[1] https://docs.google.com/document/d/1IvQJFEBFz0PnL4uVlIvt8fBS_IPwSK-avK0BRIHucxQ/edit#

vladikoff:
Fixes #483


lib/db/mysql/patches/patch-019-020.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE clients
+ADD COLUMN scopes VARCHAR(1024) AFTER trusted,
```
rfk: Naming nit: should we call this something like "allowedScopes" to make it clearer what the values here signify?

vladikoff:
@rfk r?

lib/routes/client/key_data.js
```diff
@@ -0,0 +1,113 @@
...
+      }),
+      db.getClient(Buffer.from(requestedClientId, 'hex')).then((client) => {
+        if (client && client.allowedScopes) {
+          return db.getScope(requestedScope).then((scopeDetails) => {
+            if (! scopeDetails) {
+              logger.info('keyDataRoute.noScope', {
+                requestedScope: requestedScope
+              });
+
+              throw AppError.invalidScopes(requestedScope);
```
vladikoff: @rfk do we want a custom `errno` for each of the following errors or maybe some kind of a reason `param`? or would we rather keep that private?

lib/routes/client/key_data.js
```diff
@@ -0,0 +1,113 @@
...
+const Joi = require('joi');
+
+const AppError = require('../../error');
+const db = require('../../db');
+const logger = require('../../logging')('routes.key_data');
+const P = require('../../promise');
+const validators = require('../../validators');
+const verify = require('../../browserid');
+
+const STATIC_EPOCH_ACCESS_KEY = Buffer.alloc(32).toString('hex');
```
rfk: I'm gonna be honest, I don't like this new name that Brian suggested - reading it, I don't know what an "epoch" is supposed to be and it's not obvious how it will be used by client code.  I don't have an alternative that I really *like* though...

What do you think of "additional key material"?  It captures some of the essence of how we intend the client to use it (as an additional input to the key derivation function) without introducing any new concepts like "epoch" that might be mis-understood.

Anyway, I guess it's pretty easy for us to change the name of this thing later on since it's an internal implementation detail, so we don't need to block on it.

config/dev.json
```diff
@@ -118,5 +118,12 @@
     "keyFile": "../config/key.json",
     "key": {}
   },
-  "allowHttpRedirects": true
+  "allowHttpRedirects": true,
+  "scopes": [
+    {
+      "scope": "https://identity.mozilla.org/apps/notes",
+      "description": "Read and write Notes data",
```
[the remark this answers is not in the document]
vladikoff: Yeah, we would need start extracting strings from the OAuth server once we have some use for the description. I'm wondering if I should just remove it for now though

lib/routing.js
```diff
@@ -43,6 +43,11 @@ exports.routes = [
     config: require('./routes/client/get')
   },
   {
+    method: 'POST',
+    path: v('/client/{client_id}/key-data'),
```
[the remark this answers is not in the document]
vladikoff: Good point 👍  thanks

lib/routes/client/key_data.js
```diff
@@ -0,0 +1,113 @@
...
+  },
+  response: {
+    schema: {
+      description: Joi.string().required(),
+      identifier: Joi.string().required(),
+      epochAccessKey: Joi.string().required(),
+      timestamp: Joi.number().required()
+    }
+  },
+  handler: function keyDataRoute(req, reply) {
```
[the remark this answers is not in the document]
vladikoff: 👍

lib/db/mysql/index.js
```diff
@@ -628,6 +639,18 @@ MysqlStore.prototype = {
     });
   },
 
+  getScope: function getScope (scope) {
+    return this._readOne(QUERY_SCOPE_FIND, [scope]);
+  },
+
+  registerScope: function registerScope (scope) {
```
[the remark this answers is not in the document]
vladikoff: Yeah let's keep it DB, makes it easier to add new fields without going crazy with the config ( I think :) )

vladikoff:
@rfk updated!

lib/routes/key_data.js
```diff
@@ -0,0 +1,83 @@
...
+
+const AppError = require('../error');
+const db = require('../db');
+const logger = require('../logging')('routes.key_data');
+const P = require('../promise');
+const validators = require('../validators');
+const verify = require('../browserid');
+const Scope = require('../scope');
+
+const ADDITIONAL_KEY_MATERIAL = Buffer.alloc(32).toString('hex');
```
rfk: It may be worth a comment here to say that we're using a static value on purpose, and in future we can read it from the DB.  Otherwise a casual reader (e.g. the various people inside Mozilla who've been keeping one eye on this proposal) might worry that we're doing something wrong here.

lib/routes/key_data.js
```diff
@@ -0,0 +1,83 @@
...
+      })
+    ])
+  },
+  handler: function keyDataRoute(req, reply) {
+    logger.debug('keyDataRoute.start', {
+      params: req.params,
+      payload: req.payload
+    });
+
+    const requestedScopes = new Set(Scope(req.payload.scope).values());
```
rfk: I don't think you need to turn this into a `Set`; the `Scope` class contains `has` method that IIUC should do the right thing?

lib/routes/key_data.js
```diff
@@ -0,0 +1,83 @@
...
+    const requestedClientId = req.params.client_id;
+
+    P.all([
+      verify(req.payload.assertion),
+      db.getClient(Buffer.from(requestedClientId, 'hex')).then((client) => {
+        if (client) {
+          const allowedScopes = new Set(Scope(client.allowedScopes || []).values());
+          // find all requested scopes in allowed scopes
+          const scopes = new Set([...allowedScopes].filter((x) => requestedScopes.has(x)));
+          const scopeRequests = [];
```
rfk: There's a lot of casting to `Set` here that doesn't seem necessary.  Could this be done more simply with something like the following?

```
const requestedScopes = Scope(req.payload.scope)
...
if (client) {
  const scopeRequests = [];
  const allowedScopes = Scope(client.allowedScopes)
  requestedScopes.values().forEach((s) => {
    if (allowedScopes.has(s)) {
      scopeRequests.push(db.getScope(s))
    }
  })
}
```

We could consider adding more helpers to the `Scope` class to make this easier, e.g. `intersect` and/or `filter` and/or ` forEach`.  But that can be for another day.

lib/routing.js
```diff
@@ -43,6 +43,11 @@ exports.routes = [
     config: require('./routes/client/get')
   },
   {
+    method: 'POST',
+    path: v('/client/key-data/{client_id}'),
+    config: require('./routes/key_data')
```
rfk: You moved the `.js` route implementation out from under the `clients` directory, but it's still namespaced under "/client/" in the API, which seems like a mis-match.  Should it be e.g. just "/key-data" and accept the the "client_id" as an additional parameter in the `POST` body?

test/db/index.js
```diff
@@ -565,6 +565,35 @@ describe('db', function() {
...
+    it('can register and fetch scopes', () => {
+      const scopeName = 'https://some-scope.mozilla.org';
+      const newScope = {
+        scope: scopeName,
+        hasScopedKeys: true
+      };
+      return db.registerScope(newScope)
+        .then(
+          function() {
+            return db.getScope(scopeName);
```
rfk: What should this do if the named scope does not exist?  Whatever the answer, we should probably have an explicit test for it.

lib/routing.js
vladikoff: > Should it be e.g. just "/key-data" and accept the the "client_id" as an additional parameter in the POST body?

Yeah I guess after moving it a bit that makes sense 👍

lib/routes/key_data.js
vladikoff: Good catch, thanks! 👍

test/db/index.js
vladikoff: Yeah, same as .getClient it resolves with `undefined`

PR #490, opened 2017-10-02
feat(scopes): allow https:// scopes

### LINKED ISSUE #489: Support https:// scopes
The validator for scopes does not allow https:// scopes to be requested, this needs to be fixed up:

https://github.com/mozilla/fxa-oauth-server/blob/master/lib/validators.js#L28-L30

vladikoff:
Fixes #489 

@rfk r?

### DIFF
--- a/lib/validators.js
+++ b/lib/validators.js
@@ -27,7 +27,7 @@ exports.token = Joi.string()
 
 exports.scope = Joi.string()
   .max(256)
-  .regex(/^[a-zA-Z0-9 _:]+$/);
+  .regex(/^[a-zA-Z0-9 _\/.:]+$/);
 
 // taken from mozilla/persona/lib/validate.js
 exports.assertion = Joi.string()
--- a/test/api.js
+++ b/test/api.js
@@ -344,6 +344,23 @@ describe('/v1', function() {
           assertSecurityHeaders(res);
         });
       });
+
+      it('should succeed with https:// scopes', function() {
+        const scopes = 'profile:email profile:uid https://identity.mozilla.org/apps/notes https://identity.mozilla.org/apps/lockbox';
+        const client = clientByName('Mocha');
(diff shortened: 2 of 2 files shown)


test/api.js
```diff
@@ -343,6 +343,23 @@ describe('/v1', function() {
           assertSecurityHeaders(res);
         });
       });
+
+      it('should succeed with https:// scopes', function() {
+        const scopes = 'profile:email profile:uid https://identity.mozilla.org/apps/notes https://identity.mozilla.org/apps/lockbox';
```
vladikoff: fail test, sec

vladikoff: nvm

lib/validators.js
```diff
@@ -27,7 +27,7 @@ exports.token = Joi.string()
 
 exports.scope = Joi.string()
   .max(256)
-  .regex(/^[a-zA-Z0-9 _:]+$/);
+  .regex(/^[a-zA-Z0-9 _\/.:]+$/);
```
rfk: Seeing this change made we realize that this could interact badly with or logic for checking "profile:write" scopes here:

  https://github.com/mozilla/fxa-oauth-server/blob/master/lib/scope.js#L35

rfk: In particular the "this was a weird scope. don't try to fix it, just say NO!" part of that logic.

vladikoff: @rfk are you worried about something like this `https://identity:write.mozilla.org/apps/notes:write` or something else?

vladikoff:
@rfk I think if we can ship this in train-97 then we won't need a custom branch of OAuth for Scoped Keys anymore.

lib/validators.js
rfk: I didn't have a specific concern in mind, just that the ":" in "https://" might cause weird edge-cases here that we have to watch out for.

PR #500, opened 2017-11-14
refactor(keys): rename keyMaterial, timestamp to keyRotationSecret, k…

vladikoff:
…eyRotationTimestamp

@rfk r?

### DIFF
--- a/lib/routes/key_data.js
+++ b/lib/routes/key_data.js
@@ -30,8 +30,8 @@ module.exports = {
     schema: Joi.object().pattern(/^/, [
       Joi.object({
         identifier: Joi.string().required(),
-        keyMaterial: Joi.string().required(),
-        timestamp: Joi.number().required()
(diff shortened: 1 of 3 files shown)


lib/routes/key_data.js
```diff
@@ -74,8 +74,8 @@ module.exports = {
       scopeResults.forEach((keyScope) => {
         response[keyScope.scope] = {
           identifier: keyScope.scope,
-          keyMaterial: ADDITIONAL_KEY_MATERIAL,
-          timestamp: assertionData['fxa-generation']
+          keyRotationSecret: ADDITIONAL_KEY_MATERIAL,
```
rfk: It's probably worth renaming `ADDITIONAL_KEY_MATERIAL` global for consistency.

vladikoff: Good catch!

PR #502, opened 2017-11-23
feat(keys): Check lastAuthAt freshness when fetching key data.

rfk:
For a future world where the `keyRotationSecret` is actually a secret value, we should only return it to users who have proved that they *recently* know the account password.  @vladikoff r?

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -127,6 +127,12 @@ const conf = convict({
       format: 'duration',
       default: '15 minutes',
       env: 'FXA_EXPIRATION_CODE'
+    },
+    keyDataAuth: {
+      doc: 'Key data can only be fetched if lastAuthAt is within this duration',
+      format: 'duration',
+      default: '1 hour',
+      env: 'FXA_EXPIRATION_KEY_DATA_AUTH'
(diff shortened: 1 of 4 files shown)


lib/config.js
```diff
@@ -127,6 +127,12 @@ const conf = convict({
       format: 'duration',
       default: '15 minutes',
       env: 'FXA_EXPIRATION_CODE'
+    },
+    keyDataAuth: {
+      doc: 'Key data can only be fetched if lastAuthAt is within this duration',
+      format: 'duration',
+      default: '1 hour',
+      env: 'FXA_EXPIRATION_KEY_DATA_AUTH'
```
vladikoff: 👍

lib/routes/key_data.js
```diff
@@ -71,6 +74,10 @@ module.exports = {
       const scopeResults = results[1];
       const response = {};
 
+      if (assertionData['fxa-lastAuthAt'] < Date.now() - AUTH_EXPIRES_AFTER_MS) {
```
vladikoff: 👍

PR #506, opened 2017-11-24
feat(keys): Check lastAuthAt freshness when fetching key data.

rfk:
'cause I want to deploy this for debugging....


rfk:
Turns out that `fxa-lastAuthAt` is a timestamp in seconds, not milliseconds; this is running on https://rfkelly.dev.lcip.org and appears to work correctly. @vladikoff r?

vladikoff:
from mtg: @vladikoff to check...

vladikoff:
Tested this locally with Notes and it worked, woohoo!

PR #511, opened 2018-01-11
fix(config): mark config sentryDsn and mysql password sensitive

jrgm:
r? - @vladikoff 

mark config sentryDsn, openid keys and mysql password as `sensitive`.


vladikoff:
👍

PR #514, opened 2018-01-23
feat(auth): Accept client credentials in the Authorization header.

### LINKED ISSUE #513: /v1/token does not seem to support HTTP Basic authentication scheme
I was using the client-oauth2 node library [1] to implement a client, and noticed failure in the token-obtaining step: "Err: Bad Request". This was because the library's getToken() API does not include client_id and client_secret in the request body if client_secret is available: it uses an Authorization header in the HTTP request.

I did some digging and the RFC seems to mandate that the server support this HTTP header for authentication [2].

On the other hand, it also specifies that the server may support including the parameters in the request body, but does not recommend it. Based on this I submitted a pull request [3] to client-oauth2 to support this with explicit opt-in by the caller.

I can try submitting a pull request to modify token.js around line 177 [4] if I find time later, but thought I'd open this issue in the meantime.

Please let me know if I'm missing something and/or if there's something I can read to learn more about this.

[1] https://github.com/mulesoft/js-client-oauth2
[2] https://tools.ietf.org/html/rfc6749#section-2.3.1
[3] https://github.com/mulesoft/js-client-oauth2/pull/66
[4] https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/token.js#L177

rfk:
This is required for strict compliance with RFC6749, which mandates that we MUST support client credentials via this method.

Fixes #513; @mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/token.js
+++ b/lib/routes/token.js
@@ -46,11 +46,17 @@ const ID_TOKEN_KEY = JwTool.JWK.fromObject(config.get('openid.key'), {
 
 const REFRESH_LAST_USED_AT_UPDATE_AFTER_MS = config.get('refreshToken.updateAfter');
 
+const BASIC_AUTH_REGEX = /^Basic\s+([a-z0-9+\/]+)$/i;
+
 const PAYLOAD_SCHEMA = Joi.object({
 
   client_id: validators.clientId
     .when('grant_type', {
       is: GRANT_JWT,
+      then: Joi.forbidden()
+    })
+    .when('$headers.authorization', {
+      is: Joi.string().required(),
       then: Joi.forbidden()
     }),
 
@@ -66,6 +72,10 @@ const PAYLOAD_SCHEMA = Joi.object({
     .when('grant_type', {
       is: GRANT_REFRESH_TOKEN,
       then: Joi.optional()
+    })
+    .when('$headers.authorization', {
+      is: Joi.string().required(),
+      then: Joi.forbidden()
     }),
 
   code_verifier: validators.codeVerifier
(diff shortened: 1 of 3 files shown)


lib/routes/token.js
```diff
@@ -138,19 +148,14 @@ const PAYLOAD_SCHEMA = Joi.object({
...
-    payload: function validatePayload(value, options, next) {
-      return Joi.validate(value, PAYLOAD_SCHEMA, { stripUnknown: true }, function(err, value) {
-        if (err) {
-          logger.info('routes.token.payload', err);
-        }
-
-        next(err, value);
-      });
-    }
+    payload: PAYLOAD_SCHEMA.options({ stripUnknown: true })
```
rfk: I had to simplify this so that Hapi would invoke Joi with the correct context object to make the `$headers.authorization` reference work.  We lose logging of invalid payloads but I don't think we've been looking at those anyway.

test/routes/token.js
```diff
@@ -25,8 +26,12 @@ function joiNotAllowed(err, param) {
...
 describe('/token POST', function () {
   // route validation function
-  function v(req, cb) {
-    route.validate.payload(req, {}, cb);
+  function v(req, ctx, cb) {
+    if (typeof ctx === 'function' && ! cb) {
+      cb = ctx;
+      ctx = undefined;
+    }
+    Joi.validate(req, route.validate.payload, {context: ctx}, cb);
```
rfk: This copies what hapi does to pass around context internally, but I made sure to also add some live API tests covering the `$headers.authorization` reference stuff, in case it breaks in future.

test/routes/token.js
```diff
@@ -139,17 +170,6 @@ describe('/token POST', function () {
...
-      v({
-        client_secret: CLIENT_SECRET,
-        code: CODE,
-        grant_type: GRANT_JWT,
-      }, (err) => {
-        joiNotAllowed(err, 'client_secret');
-        done();
-      });
-    });
-
```
rfk: This test was an exact duplicate of the one immediately above it, so I took the opportunity to remove it.

lib/routes/token.js
```diff
@@ -46,12 +46,18 @@ const ID_TOKEN_KEY = JwTool.JWK.fromObject(config.get('openid.key'), {
 
 const REFRESH_LAST_USED_AT_UPDATE_AFTER_MS = config.get('refreshToken.updateAfter');
 
+const BASIC_AUTH_REGEX = /^Basic\s+([A-Za-z0-9+\/]+)$/i;
```
philbooth: Minor nit, but this regex sets the ignore-case flag then goes to the effort of including both cases in the character class. Presumably one or other of those is unnecessary?

rfk: Indeed, nice catch!  I used `/i` for matching the "Basic" part but didn't consider the rest.

PR #522, opened 2018-02-26
chore(npm): update to npm5

vladikoff:


npm-shrinkwrap.json
```diff
@@ -1,7756 +1,8174 @@
 {
   "name": "fxa-oauth-server",
   "version": "1.106.0",
+  "lockfileVersion": 1,
```
vladikoff: I had an idea to parse and check to make sure new commits have `"lockfileVersion": 1,` (this will check to make sure the new wrap was created with npm5) but we might not need this

vladikoff:
@mozilla/fxa-devs @vbudhram  r?

PR #530, opened 2018-03-15
feat(amr): Report `amr` and `acr` claims in the id_token.

### LINKED ISSUE #519: Report MFA status to reliers at the end of the OAuth dance
To make informed security decisions, reliers need to know whether an OAuth flow was completed using MFA and may want to inspect the details of what kind of MFA took place.  We need to decide on an API for exposing this data, most likely as part of the OIDC "id_token".

First action here is for @rfk to review OIDC and related specs for prior art on this topic.

rfk:
Fixes #519.  Work-in-progress for now while I try to get all the servers stood up and talking to each other in a dev environment.

### DIFF
--- a/lib/browserid.js
+++ b/lib/browserid.js
@@ -3,12 +3,24 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 
+const Joi = require('joi');
+
 const AppError = require('./error');
 const config = require('./config');
 const logger = require('./logging')('assertion');
 const P = require('./promise');
 
 const HEX_STRING = /^[0-9a-f]+$/;
+const CLAIMS_SCHEMA = Joi.object({
+  'uid': Joi.string().length(32).regex(HEX_STRING).required(),
+  'fxa-generation': Joi.number().integer().min(0).required(),
+  'fxa-verifiedEmail': Joi.string().max(255).required(),
+  'fxa-lastAuthAt': Joi.number().integer().min(0).required(),
+  'fxa-tokenVerified': Joi.boolean().optional(),
+  'fxa-amr': Joi.array().items(Joi.string().alphanum()).optional(),
+  'fxa-aal': Joi.number().integer().min(0).max(3).optional(),
+}).options({ stripUnknown: true });
+
 const AUDIENCE = config.get('publicUrl');
 
 const request = require('request').defaults({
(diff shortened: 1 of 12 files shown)


lib/browserid.js
```diff
@@ -68,28 +80,25 @@ module.exports = function verifyAssertion(assertion) {
...
+    const allowedIssuer = config.get('browserid.issuer');
     if (parts.length !== 2 || parts[1] !== allowedIssuer) {
       return error('invalid email', email);
     }
     if (body.issuer !== allowedIssuer) {
       return error('invalid issuer', body.issuer);
     }
-    var uid = parts[0];
-    //sanity check the uid
-    if (uid.length !== 32 && HEX_STRING.test(uid)) {
```
rfk: I'm pretty sure this old validation code was incorrect, and should have been `|| ! HEX_STRING.test(uid)`.  I've added some tests for the validation here, and moved it into a Joi schema so it's more declarative.

lib/browserid.js
```diff
@@ -3,12 +3,24 @@
...
 const P = require('./promise');
 
 const HEX_STRING = /^[0-9a-f]+$/;
+const CLAIMS_SCHEMA = Joi.object({
+  'uid': Joi.string().length(32).regex(HEX_STRING).required(),
+  'fxa-generation': Joi.number().integer().min(0).required(),
+  'fxa-verifiedEmail': Joi.string().max(255).required(),
+  'fxa-lastAuthAt': Joi.number().integer().min(0).required(),
+  'fxa-tokenVerified': Joi.boolean().optional(),
+  'fxa-amr': Joi.array().items(Joi.string().alphanum()).optional(),
```
vbudhram: Should be a little more strict and allow only `email, pwd, otp` here?

lib/browserid.js
```diff
@@ -3,12 +3,24 @@
...
+  'fxa-aal': Joi.number().integer().min(0).optional(),
```
vbudhram: Could also set a max of `3`.

lib/browserid.js
rfk: I'm in two minds about this.  It seems worthwhile for sanity-checking, but the oauth-server doesn't actually interpret this values at all, it just passes them on to the reliers.  If we ever add another method, do we want to require a coordinated oauth-server deploy, or just get it "for free" when the auth-server updates?  I lean towards the latter.

vbudhram: > oauth-server doesn't actually interpret this values at all

Consider me convinced!

PR #533, opened 2018-03-20
fix(route): make email false by default

vladikoff:


lib/routes/verify.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
   validate: {
     payload: {
       token: validators.token.required(),
-      email: Joi.boolean().default(true)
+      email: Joi.boolean().default(false)
```
vladikoff: @rfk hm so current version of the PR will stop sending the email unless `email:true` is specified, might break something?

rfk: Right.  Unfortunately I don't see a good way around this.  Any reliers that still expect the email here *should* be explicitly requesting it rather than relying on the default value here, but how can we check that without breaking any that aren't?  I think they've all been updated but it's hard to be sure.

Perhaps we should aim to do this as a point-release so we can carefully QA it (perhaps with that other breaking OAuth change, if we don't manage to confirm it from request logs?)

rfk:
>  I think they've all been updated but it's hard to be sure.

Actually, I've been thinking about this wrong.  The change is on the `/v1/verify` endpoint, which is not used by OAuth reliers like AMO or Pontoon.  It's used by OAuth service providers like fxa-basket-proxy and fxa-profile-server and Kinto.  I'm fairly confident our own SPs have been updated, and Kinto has no reason to be using it, so the potential fallout seems lower here.

vladikoff:
> Perhaps we should aim to do this as a point-release so we can carefully QA it (perhaps with that other breaking OAuth change, if we don't manage to confirm it from request logs?)

This is for train-109, I will add a TODO to check Kinto, let me know if there are other ones

vladikoff:
@rfk and to finally fix the issue we can remove the email boolean in train-110 maybe?

rfk:
>  let me know if there are other ones

I can't think of any TBH, at least not any that are so old that they'd be depending on the email being there.

> @rfk and to finally fix the issue we can remove the email boolean in train-110 maybe?

👍, assuming that nobody is actually sending it with `email: false` :-)

PR #538, opened 2018-04-05
fix(mock): Fixes #334 Mock outstanding error logs in test suite

### LINKED ISSUE #334: Mock outstanding error logs in test suite
It's currently noisy running the tests, as when we test some error conditions of the database, the errors are shot out to console. Since these errors are expected, it'd be better to mock them out, so the test suite doesn't look like something horrible happened.

deeptibaghel:
Fixes #334 Mock outstanding error logs in test suite.
Also handles mozlog deprecation warning.


vladikoff:
Thanks @deeptibaghel !!

PR #543, opened 2018-04-18
refactor(email): Fixes #352 Remove ability to fetch email address 

### LINKED ISSUE #352: Remove ability to fetch email address from the oauth-server API

deeptibaghel:
@vladikoff we need to remove just this parameter or drop from the response also  :
https://github.com/mozilla/fxa-oauth-server/blob/7ad1e56f95ec54dd5096937a5cb3b42585841105/lib/routes/verify.js#L23

Also this will remain as it is ?
https://github.com/mozilla/fxa-oauth-server/blob/7ad1e56f95ec54dd5096937a5cb3b42585841105/lib/token.js#L60

### DIFF
--- a/lib/routes/verify.js
+++ b/lib/routes/verify.js
@@ -11,7 +11,8 @@ const validators = require('../validators');
 module.exports = {
   validate: {
     payload: {
-      token: validators.token.required()
+      token: validators.token.required(),
+      email: Joi.boolean().default(false)
     }
   },
   response: {
--- a/lib/routes/verify.js
+++ b/lib/routes/verify.js
@@ -11,8 +11,7 @@ const validators = require('../validators');
 module.exports = {
   validate: {
     payload: {
-      token: validators.token.required(),
-      email: Joi.boolean().default(false)
+      token: validators.token.required()
     }
   },
   response: {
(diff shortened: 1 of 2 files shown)


test/api.js
```diff
@@ -2768,7 +2768,7 @@ describe('/v1', function() {
       });
     });
 
-    it('should not return the email if email:false for profile:email scope', function() {
+    it('fails for payload having email:false', function() {
```
shane-tomlinson: Because we have to allow `email: false` for one train until all services stop sending it, these test updates need to wait one train.

lib/routes/verify.js
```diff
@@ -25,15 +25,7 @@ module.exports = {
   },
   handler: function verify(req, reply) {
     token.verify(req.payload.token).then(function(info) {
-      if (req.payload.email) {
```
rfk: I think it would be worth making `email` be `Joi.boolean().optional()` in the payload schema above, and changing this line to log a warning whenever `req.payload.email !== undefined`.  This would help us confirm that all services have indeed stopped sending an explicit `email: false` in their requests.

deeptibaghel: @rfk done

PR #548, opened 2018-04-27
fix(validation): Allow redirect uris with existing query params.

### LINKED ISSUE #547: Allow use of query strings in redirect_uri
The validation introduced in #534 make it an error.

rfk:
A quick fix for #547; we don't do anything with these URIs other than tunnel them around to the content-server, which does its own additional validation.  @mozilla/fxa-devs r?

Fixes #547.


l-hedgehog:
Thanks!

PR #551, opened 2018-05-02
fix(scopes): Formalize and document scope-handling rules.

rfk:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/575

While working on adding OAuth support for sync, I figured we should have a canonical reference for what sorts of scope we support and how we check for matches.  So I started something, and here's the rendered view:

  https://github.com/mozilla/fxa-oauth-server/blob/scopes-documentation/docs/scopes.md

I want to add some explicit testcases for this in the code here, but am putting the doc up for early feedback.  @vladikoff does this seem sane?  I think it differs slightly from the proposal in the original scoped-keys doc, in that it uses the URL fragment for qualifications like "readonly".  But I quite like the ability to check scopes by parsing them as URL resource references.


docs/scopes.md
```diff
@@ -0,0 +1,179 @@
...
+* `https://identity.mozilla.com/apps/oldsync#read` indicates
+   read-only access to the user's data in Firefox Sync.
+* `https://identity.mozilla.com/apps/oldsync/history#write` indicates
+   write-only access to the user's history data in Firefox Sync.
+
+To be a valid scope, the URL must:
+
+* Be an absolute `https://` URL.
+* Have no query component.
+* Have a fragment component consisting only of alphanumeric ascii characters.
```
philbooth: Random drive-by comment, but this wording makes it sound like the fragment part is mandatory because of the MUST before the list. Maybe clarify here with an "if specified..." clause, or split out to a separate MAY list?

philbooth: (or include MUST / MAY in each bullet rather than aggregated at the top)

rfk: Thanks, good call, I will update.

docs/scopes.md
```diff
@@ -0,0 +1,201 @@
...
+
+FxA supports a small set of "short-name" scope values
+that are identified by a short English word.
+These correspond either to
+scope values defined by external specifications (e.g. OpenID Connect),
+or to legacy scope values
+introduced during earily development.
+
+**No new short-name scope values should be added.**
+Instead we prefer to use URLs for new scope values,
```
vladikoff: > +**No new short-name scope values should be added.**
> Instead we prefer to use URLs for new scope values,

Ah interesting 👍 . I'm surprised to hear this but i guess this makes sense after looking at Google's list of these scopes

docs/scopes.md
```diff
@@ -0,0 +1,201 @@
...
+
+* `https://identity.mozilla.com/apps/oldsync` indicates full
+   access to the user's data in Firefox Sync.
+* `https://identity.mozilla.com/apps/oldsync/bookmarks` indicates
+   full access to the user's bookmark data in Firefox Sync,
+   but not to other data types.
+* `https://identity.mozilla.com/apps/oldsync#read` indicates
+   read-only access to the user's data in Firefox Sync.
+* `https://identity.mozilla.com/apps/oldsync/history#write` indicates
+   write-only access to the user's history data in Firefox Sync.
```
vladikoff: yeah!!

rfk:
Thanks @vladikoff; I think I'll leave this open until I have a corresponding PR to use it in some of our other codebases, in case that reveals some opportunities for API cleanup.

shane-tomlinson:
@rfk - is there anything else you want to add to this?

GitCop:

 * Commit: 5d198fc378df369efd1dd8adfb1082851a1a15a3

---

GitCop:

 * Commit: 5d198fc378df369efd1dd8adfb1082851a1a15a3

---

GitCop:

 * Commit: 5d198fc378df369efd1dd8adfb1082851a1a15a3

---

GitCop:

 * Commit: 5d198fc378df369efd1dd8adfb1082851a1a15a3

---

shane-tomlinson:
@rfk - what's left to do here?

rfk:
This has been updated to use a tag of fxa-shared rather than a github ref; @mozilla/fxa-devs r?

docs/scopes.md
```diff
@@ -0,0 +1,199 @@
...
+scope-checking code can be validated.
+
+Valid implications:
+* `profile:write` implies `profile`.
+* `profile` implies `profile:email`.
+* `profile:write` implies `profile:email`.
+* `profile:write` implies `profile:email:write`.
+* `profile:email:write` implies `profile:email`.
+* `profile profile:email:write` implies `profile:email`.
+* `profile profile:email:write` implies `profile:display_name`.
```
vbudhram: Did you mean `profile profile:display_name:write` implies `profile:display_name`?

docs/scopes.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Profile data
+
+* `profile`: access the user's profile data.
+* `profile:uid`: access the user's opaque user id.
+* `profile:email`: access the user's email address.
+* `profile:locale`: access the user's locale.
+* `profile:avatar`: access the user's avatar picture.
+* `profile:display_name`: access the user's human-readable display name.
+* `profile:amr`: access information about the user's authentication methods and 2FA status.
```
shane-tomlinson: Should there be any indication on whether any of these are read vs read/write? uid/locale/amr I imagine are read-only.

docs/scopes.md
```diff
@@ -0,0 +1,199 @@
...
+For example:
+
+* `https://identity.mozilla.com/apps/oldsync` indicates full
+   access to the user's data in Firefox Sync.
+* `https://identity.mozilla.com/apps/oldsync/bookmarks` indicates
+   full access to the user's bookmark data in Firefox Sync,
+   but not to other data types.
+* `https://identity.mozilla.com/apps/oldsync#read` indicates
+   read-only access to the user's data in Firefox Sync.
+* `https://identity.mozilla.com/apps/oldsync/history#write` indicates
```
shane-tomlinson: Why would one want write-only access to history?

docs/scopes.md
```diff
@@ -0,0 +1,199 @@
...
+
+* `https://identity.mozilla.com/apps/oldsync`: access to data in Firefox Sync.
+* `https://identity.mozilla.com/apps/notes`: access to data in Firefox Notes.
+
+
+## Scope Matching and Implication
+
+We say that a scope value A *implies* another scope value B
+if they are exactly equal,
+of if A represents a more general capability than B.
```
shane-tomlinson: `of if`=> `or if`

docs/scopes.md
shane-tomlinson: > Did you mean profile profile:display_name:write implies profile:display_name?

`profile` implies `profile:display_name` (both read-only), the `profile:email:write` is needed to write to the email.

docs/scopes.md
```diff
@@ -0,0 +1,199 @@
...
+Invalid implications:
+* `profile:email:write` does *not* imply `profile`.
+* `profile:email:write` does *not* imply `profile:write`.
+* `profile:email` does *not* imply `profile:display_name`.
+* `profilebogey` does *not* imply `profile`.
+* `profile:write` does *not* imply `https://identity.mozilla.com/apps/oldsync`.
+* `profile profile:email:write` does *not* imply `profile:write`.
+* `https` does *not* imply `https://identity.mozilla.com/apps/oldsync`.
+* `https://identity.mozilla.com/apps/oldsync` does *not* imply `profile`.
+* `https://identity.mozilla.com/apps/oldsync#read` does *not* imply `https://identity.mozila.com/apps/oldsync/bookmarks`.
```
shane-tomlinson: In the URL after "does not imply" do you want "mozila" or "mozilla"? Applies to this and the following 6 lines.

vladikoff: should be `mozilla ` yeap :)

docs/scopes.md
rfk: It's true that the `amr` cannot be modified, but "profile:amr:write" is still a well-formed scope with well-defined semantics (that doesn't happen to grant you any addition powers over "profile:amr".

docs/scopes.md
rfk: If you're a mobile app that wants to record visits that can show up in the activity-stream on the user's desktop, but we don't want to let you read the sites that the user has been visiting on other devices.

rfk:
Thanks @vbudhram @shane-tomlinson, nits fixed.

PR #561, opened 2018-06-07
feat(authorization): Require tokenVerified=true when granting key-bearing scopes.

rfk:
The sync tokenserver does a special check for "fxa-tokenVerified" in order to enforce the use of session verification when accessing sync:

  https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L140

Let's apply the same check here before granting any scopes that come with keys.  In theory the user should *always* have a verified assertion when requesting one of these scopes, because they will have just done a keyfetch that would have required it.  But there is at least one known series of calls to our backend that can yield keys without doing a verification, so it makes sense to double-check here and avoid any loopholes.

@vladikoff r?

### DIFF
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -213,6 +213,28 @@ module.exports = {
           exitEarly = true;
           throw AppError.invalidAssertion();
         }
+        // Any request for a key-bearing scope should be using a verified token.
+        // Double-check that here as a defense-in-depth measure.
+        if (! claims['fxa-tokenVerified']) {
+          return P.each(scope.values(), scope => {
+            // Don't bother hitting the DB if other checks have failed.
+            if (exitEarly) {
+              return;
+            }
+            // We know only URL-format scopes can have keys,
+            // so avoid trips to the DB for common scopes like 'profile'.
+            if (scope.startsWith('https://')) {
+              return db.getScope(scope).then(s => {
+                if (s.hasScopedKeys) {
+                  exitEarly = true;
+                  throw AppError.invalidAssertion();
+                }
+              });
+            }
+          }).then(() => {
+            return claims;
+          });
+        }
         return claims;
       }),
       db.getClient(Buffer.from(req.payload.client_id, 'hex')).then(function(client) {
@@ -272,6 +294,9 @@ module.exports = {
         req.payload.redirect_uri = uri;
 
         return client;
+      }).catch(err => {
+        exitEarly = true;
+        throw err;
       }),
       scope.values(),
       req
--- a/lib/routes/key_data.js
+++ b/lib/routes/key_data.js
@@ -26,7 +26,7 @@ module.exports = {
     payload: {
       client_id: validators.clientId,
       assertion: validators.assertion.required(),
-      scope: Joi.string()
+      scope: validators.scope
     }
   },
   response: {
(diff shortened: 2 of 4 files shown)


lib/routes/authorization.js
```diff
@@ -213,6 +213,23 @@ module.exports = {
...
+        // Any request for a key-bearing scope should be using a verified token.
+        // Double-check that here as a defense-in-depth measure.
+        if (! claims['fxa-tokenVerified']) {
+          return P.each(scope.values(), scope => {
+            // We know only URL-format scopes can have keys,
+            // so avoid trips to the DB for common scopes like 'profile'.
+            if (scope.startsWith('https://')) {
+              return db.getScope(scope).then(s => {
+                if (s.hasScopedKeys) {
+                  throw AppError.invalidAssertion();
```
shane-tomlinson: Do you need an `exitEarly = true` [like up here](https://github.com/mozilla/fxa-oauth-server/pull/561/files#diff-422a5c70fd5a58a549ee0f442ed2e452R213)?

rfk: Hmm...thinking about it, I wonder if that earlier `exitEarly` ever actually triggers.  Verifying assertions is pretty slow because of public-key crypto, and hitting the database is pretty fast, it wouldn't surprise me if the second are almost always wins this race.  I'm happy to add it but I'm not confident it'd have any effect in practice.

lib/routes/authorization.js
```diff
@@ -213,6 +213,23 @@ module.exports = {
           exitEarly = true;
           throw AppError.invalidAssertion();
         }
+        // Any request for a key-bearing scope should be using a verified token.
+        // Double-check that here as a defense-in-depth measure.
+        if (! claims['fxa-tokenVerified']) {
+          return P.each(scope.values(), scope => {
```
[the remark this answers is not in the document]
rfk: Good call.  We expect the set of supported scopes and their properties to change very rarely, it'd probably be better to refactor this to keep them in memory and periodically refresh from the db.

lib/routes/authorization.js
rfk: I went ahead and added this, and also added a *check* in this first promise chain, in case the second promise chain is the one that finishes first.

lib/routes/authorization.js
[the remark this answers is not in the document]
rfk: We don't currently have any clients that request more than one url-style scope, so I'm not too worried about it for now, but I did file a follow-up issue https://github.com/mozilla/fxa-oauth-server/issues/565

rfk:
@vladikoff @jrgm what do you think about this as a potential point-release for train-113?  It's currently the only outstanding blocker for re-enabling [OAuth support in the sync tokenserver](https://bugzilla.mozilla.org/show_bug.cgi?id=1455219) in production.

lib/routes/authorization.js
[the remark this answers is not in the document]
vladikoff: 👍

lib/routes/key_data.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
     payload: {
       client_id: validators.clientId,
       assertion: validators.assertion.required(),
-      scope: Joi.string()
+      scope: validators.scope
```
vladikoff: 👍

PR #562, opened 2018-06-08
feat(ci): add python for native newrelic metrics

### LINKED ISSUE #555: Add python for native metrics in Dockerfile
Just like this: https://github.com/mozilla/fxa-auth-server/blob/master/Dockerfile-build#L3

vladikoff:
Fixes #555

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,6 +1,6 @@
 FROM node:8.11.1-alpine
 
-RUN apk add --no-cache git
+RUN apk add --no-cache git make gcc g++ linux-headers openssl python
 
 RUN addgroup -g 10001 app && \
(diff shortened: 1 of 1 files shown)


Dockerfile-build
```diff
@@ -1,6 +1,6 @@
 FROM node:8.11.1-alpine
 
-RUN apk add --no-cache git
+RUN apk add --no-cache git make gcc g++ linux-headers openssl python
```
rfk: Sounds fine to me, but do we want to uninstall some of the build deps at the end in order to reduce the footprint of the final image?

vladikoff: moved to https://github.com/mozilla/fxa-oauth-server/pull/563

PR #563, opened 2018-06-08
feat(ci): add python for native newrelic metrics

### LINKED ISSUE #555: Add python for native metrics in Dockerfile

vladikoff:
Fixes #555

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,6 +1,6 @@
 FROM node:8.11.1-alpine
 
-RUN apk add --no-cache git
+RUN apk add --no-cache git make gcc g++ linux-headers openssl python
 
 RUN addgroup -g 10001 app && \
     adduser -D -G app -h /app -u 10001 app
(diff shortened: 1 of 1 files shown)


Dockerfile-build
```diff
@@ -20,3 +20,7 @@ COPY scripts/gen_keys.js scripts/gen_keys.js
 RUN npm install --production && rm -rf ~app/.npm /tmp/*
 
 COPY . /app
+
+USER root
+
+RUN apk del make gcc g++ linux-headers openssl python
```
vladikoff: Even though i did this the image stayed at over 100 mb 

 mystery

vladikoff: Ref: https://github.com/mozilla/fxa-oauth-server/pull/562#discussion_r193951641

vladikoff: ah yes because of this https://github.com/gliderlabs/docker-alpine/issues/45

PR #566, opened 2018-06-09
fix(pkce): Don't require PKCE in the direct grant flow.

### LINKED ISSUE #559: Don't require PKCE for the direct grant flow
We currently require public clients to provide a PKCE challenge regardless of the grant type they're requesting.  If a publicClient is requesting a direct grant with an assertion, it shouldn't need to provide PKE challenge.

Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1460646#c7

rfk:
Fixes #559.

### DIFF
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -119,6 +119,31 @@ function generateGrant(claims, client, scope, req) {
   });
 }
 
+// Check that PKCE is provided if and only if appropriate.
+function checkPKCEParams(req, client) {
+  if (req.payload.response_type === TOKEN) {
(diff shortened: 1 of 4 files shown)


lib/routes/authorization.js
```diff
@@ -119,6 +119,31 @@ function generateGrant(claims, client, scope, req) {
   });
 }
 
+// Check that PKCE is provided if and only if appropriate.
+function checkPKCEParams(req, client) {
+  if (req.payload.response_type === TOKEN) {
+    // Direct token grant can't use PKCE.
```
vladikoff: 👍

PR #576, opened 2018-07-04
fix(mysql): Correctly aggregate tokens by clientid.

rfk:
It turns out that the mysql variant of https://github.com/mozilla/fxa-oauth-server-private/pull/9 was buggy, and would aggregate all clients into a single record with id of "undefined" rather than producing one record per client id.

It's a simple fix that we should point-release to train-115; @vladikoff r?

But it's also an indication that the CI in fxa-oauth-server-private is only running tests against the in-memory db, not against mysql.  @jrgm @jbuck which do you think would be a simpler way to fix that - having circle run tests against the mysql backend, or adding travis to the private repo?


jrgm:
> having circle run tests against the mysql backend, or adding travis to the private repo?

Not opposed to the latter, but I think running circle with DB=mysql would be good.

jrgm:
Or, against both, actually, since it's not much time to run.

PR #577, opened 2018-07-06
feat(oauth): Implement incremental authorization features.

### LINKED ISSUE #553: Support "incremental authorization"
OAuth has a nice flow called "incremental authorization" that allows a client to request scopes on an as-needed basis, and have them added to its existing refresh_token rather juggling multiple refresh_tokens:

* https://tools.ietf.org/html/draft-wdenniss-oauth-incremental-auth-01
* https://developers.google.com/identity/protocols/OAuth2WebServer#incrementalAuth

I think this will be useful as we build out a refresh-token-focussed client lib over in https://github.com/mozilla/application-services/issues/40

eoger:
Fixes #553 

This is a first draft that probably needs more work.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -244,7 +244,6 @@ AppError.notPublicClient = function unknownClient(clientId) {
   });
 };
 
-
 AppError.mismatchCodeChallenge = function mismatchCodeChallenge(pkceHashValue) {
   return new AppError({
     code: 400,
@@ -273,6 +272,17 @@ AppError.staleAuthAt = function staleAuthAt(authAt) {
     message: 'Stale authentication timestamp'
   }, {
     authAt: authAt
+  });
+};
+
+AppError.notConfidentialClient = function notConfidential(clientId) {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: 120,
+    message: 'Not a confidential client'
+  }, {
+    clientId: clientId
   });
 };
 
--- a/lib/routes/authorization.js
+++ b/lib/routes/authorization.js
@@ -63,6 +63,12 @@ function detectInvalidScopes(requestedScopes, validScopes) {
 }
 
 function generateCode(claims, client, scope, req) {
+  if (client.publicClient && req.payload.include_granted_scopes) {
+    // From RFC (https://tools.ietf.org/html/draft-ietf-oauth-incremental-authz-00):
+    // the "include_granted_scopes" feature MUST NOT be enabled for such public
+    // client requests.
+    throw AppError.notConfidentialClient();
+  }
   return db.generateCode({
     clientId: client.id,
     userId: buf(claims.uid),
@@ -75,6 +81,7 @@ function generateCode(claims, client, scope, req) {
     codeChallengeMethod: req.payload.code_challenge_method,
     codeChallenge: req.payload.code_challenge,
     keysJwe: req.payload.keys_jwe,
+    includeGrantedScopes: req.payload.include_granted_scopes,
   }).then(function(code) {
     logger.debug('redirecting', { uri: req.payload.redirect_uri });
 
@@ -194,6 +201,7 @@ module.exports = {
           then: Joi.optional(),
           otherwise: Joi.forbidden()
         }),
+      include_granted_scopes: Joi.boolean().optional(),
       keys_jwe: validators.jwe
         .when('response_type', {
           is: CODE,
--- a/lib/routes/token.js
+++ b/lib/routes/token.js
@@ -124,7 +124,10 @@ const PAYLOAD_SCHEMA = Joi.object({
     .when('grant_type', {
       is: GRANT_JWT,
       otherwise: Joi.forbidden()
-    })
+    }),
+
+  existing_grant: validators.token
+    .optional()
 
 });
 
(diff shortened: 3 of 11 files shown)


shane-tomlinson:
FWIW, here's the proposed OAuth 2.0 spec for Incremental Authorization - https://tools.ietf.org/html/draft-ietf-oauth-incremental-authz-00

lib/scope.js
```diff
@@ -50,6 +50,16 @@ Scope.prototype = {
...
 
+  merge: function merge(scope) {
+    if (! (scope instanceof Scope)) {
+      throw new Error('Scope merge only works with Scope instances.');
+    }
+    const ours = this.values();
+    const theirs = scope.values();
+    const merged = Array.from(new Set([...ours, ...theirs]));
+    return Scope(merged);
+  },
```
rfk: 👍

lib/routes/token.js
```diff
@@ -236,6 +240,32 @@ module.exports = {
...
+            throw new AppError.unauthorized('Unknown existing grant');
+          }
+          if (tok.userId.toString('hex') !== vals.userId.toString('hex') ||
+              tok.clientId.toString('hex') !== vals.clientId.toString('hex')) {
+            throw new AppError.unauthorized('Invalid grant');
+          }
+          vals.scope = Scope(vals.scope).merge(Scope(tok.scope)).values();
+          return vals;
+        });
+      } else if (vals.includeGrantedScopes) {
```
rfk: What, if anything, should happen if both `existing_grant` and `include_granted_scopes` are specified at the same time?

test/api.js
```diff
@@ -1516,6 +1516,141 @@ describe('/v1', function() {
...
+        }).then(function(res) {
+          assert.equal(res.statusCode, 200);
+          assertSecurityHeaders(res);
+          assert.equal(res.result.token_type, 'bearer');
+          assert(res.result.access_token);
+          assert(res.result.refresh_token);
+          assert.equal(res.result.access_token.length,
+            config.get('unique.token') * 2);
+          assert.equal(res.result.refresh_token.length,
+            config.get('unique.token') * 2);
```
rfk: FWIW I don't think these detailed "is the token in the expected format?" tests are necessary to duplicate here, that should be well covered by existing tests.

lib/routes/authorization.js
```diff
@@ -63,6 +63,11 @@ function detectInvalidScopes(requestedScopes, validScopes) {
 }
 
 function generateCode(claims, client, scope, req) {
+  if (client.publicClient && req.payload.include_granted_scopes) {
+    // From RFC: the "include_granted_scopes" feature MUST NOT be enabled for such public
+    // client requests.
+    throw AppError.notConfidentialClient();
```
rfk: We should add an explicit test for this as well.

test/api.js
```diff
@@ -1516,6 +1516,141 @@ describe('/v1', function() {
...
+              code: res.result.code,
+              existing_grant: firstRefreshToken,
+            }
+          });
+        }).then(function(res) {
+          assert.equal(res.statusCode, 200);
+          assertSecurityHeaders(res);
+          assert.equal(res.result.token_type, 'bearer');
+          assert(res.result.access_token);
+          assert(res.result.refresh_token);
```
rfk: IIUC, this grants a *new* refresh token that has the scopes of the old one, plus the newly-granted scopes, is that correct?  If so, we should probably explicitly test that it's a new token.

To be honest though, I wonder if it makes sense to re-use the existing refresh token in this case.  Or perhaps to immediately revoke the old refresh token when returning the new one?  We'll have to think about what the right behaviour is from the client side here...

lib/routes/authorization.js
```diff
@@ -63,6 +63,11 @@ function detectInvalidScopes(requestedScopes, validScopes) {
 }
 
 function generateCode(claims, client, scope, req) {
+  if (client.publicClient && req.payload.include_granted_scopes) {
+    // From RFC: the "include_granted_scopes" feature MUST NOT be enabled for such public
+    // client requests.
```
shane-tomlinson: Can you link to the RFC here?

lib/error.js
```diff
@@ -276,4 +275,15 @@ AppError.staleAuthAt = function staleAuthAt(authAt) {
   });
 };
 
+AppError.notConfidentialClient = function notConfidential(clientId) {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: 120,
```
shane-tomlinson: Can you document this error value [in api.md](https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md)? This error will need to be added to the content-server's [oauth-errors file](https://github.com/mozilla/fxa-content-server/blob/29c7fb8e473d53d793d85766ccaaaae89dd2a1ab/app/scripts/lib/oauth-errors.js#L90) too. It looks like someone forgot errno 119 so I [opened a bug](https://github.com/mozilla/fxa-content-server/issues/6358).

lib/routes/token.js
shane-tomlinson: From the [proposed spec](https://tools.ietf.org/html/draft-ietf-oauth-incremental-authz-00#section-4) it looks like `include_granted_scopes` is only for confidential clients whereas `existing_grant` is intended primarily for public clients but can be used for confidential clients (at least that's my interpretation of `Public clients (and confidential clients using this technique)`. It does not state what should happen if both are specified for a confidential client, my guess would be the union but I also don't understand how the scopes granted using `existing_grant` would be anything but a subset of scopes granted in `include_granted_scopes`.

Asking for clarification for this scenario from the spec authors might be useful since they are actively seeking feedback at this time.

lib/routes/token.js
```diff
@@ -236,6 +240,32 @@ module.exports = {
...
+        return db.getActiveClientTokensByUid(vals.userId.toString('hex')).then(function (tokens) {
+          const allScopes = []; // Contain duplicates, but Scope#merge drops them.
+          for (const token of tokens) {
+            allScopes.push(...token.scope);
```
shane-tomlinson: Seems strange to me that the singular `scope` is really an array, but that's not an issue caused by this PR.

lib/routes/token.js
eoger: Ping @WilliamDenniss, do you know what should happen in this case?

eoger:
> Do you have plans to start consuming this from the fxa-rust-client in the short term?

I wonder if this also needs a PR on the content-server side: "You have already given the follow permissions, do you want to add these?"

shane-tomlinson:
> I wonder if this also needs a PR on the content-server side:

We never built the frontend to handle incremental auth, so yup, the content server will need an update too. You've been so prolific lately I just assumed it'd be forthcoming. ;)

PR #578, opened 2018-07-09
feat(codes): Delete authorization codes when revoking client access.

rfk:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1320218#c3, revoking access for an OAuth client should revoke everything, including any as-yet-unclaimed authorization codes.


lib/db/mysql/index.js
```diff
@@ -474,6 +476,11 @@ MysqlStore.prototype = {
    * @returns {Promise}
    */
   deleteActiveClientTokens: function deleteActiveClientTokens(clientId, uid) {
```
philbooth: Possible bike-shedding alert, but does this function deserve a more generic name now that it also deletes codes as well as tokens, or is the distinction between the two sufficiently blurry that "tokens" works well as a collective term for both entities?

rfk: I went with 'deleteClientAuthorization' which seems a little more generic - we delete anything that authorizes the client to do things on the user's account.

PR #579, opened 2018-07-10
Dockerpush fix docker use base builder run npm ls production

jrgm:
This is similar to mozilla/fxa-profile-server#330. It builds from a base image that has tools that can build binary node extensions, and secondly, checks the install with `npm install --production`.

The net gain is 3MB from the production image. (A `--squash` would recoup about 8MB from that image, but not bothering with `--squash` in this PR).

r? - @vladikoff

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,8 +1,9 @@
-FROM node:8-alpine
+FROM node:8-alpine AS builder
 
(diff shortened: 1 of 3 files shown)


Dockerfile-build
```diff
@@ -1,8 +1,9 @@
-FROM node:8-alpine
+FROM node:8-alpine AS builder
```
vladikoff: nice! 👍

PR #580, opened 2018-07-10
fix(purge): add purgeExpiredTokensById to select, then delete by primary key

### LINKED ISSUE #571: Need a different strategy for purging expired oauth tokens
The problem is that the current bin/purge-expired-tokens.js is 1) very slow, and 2) causes some number of failed oauth and profile api calls on timeouts.

So, perhaps something like a) selecting ~100000 rows in a date range, filitering out the pocket ids in memory, and then batching out DELETE FROM tokens where token in (...) and expires < now(), or b) dumping out all candidate tokens to disk (or queue) and batching from that list.

jrgm:
r? - @mozilla/fxa-devs 

(Hopefully*) fixes #571 

[*] it does what I intended it to do, but how it plays in performance/non-contention in a huge production table is TBD.


lib/db/mysql/index.js
```diff
@@ -189,6 +190,12 @@ const QUERY_CODE_DELETE_USER = 'DELETE FROM codes WHERE userId=?';
 const QUERY_DEVELOPER = 'SELECT * FROM developers WHERE email=?';
 const QUERY_DEVELOPER_DELETE = 'DELETE FROM developers WHERE email=?';
 const QUERY_PURGE_EXPIRED_TOKENS = 'DELETE FROM tokens WHERE clientId NOT IN (?) AND expiresAt < NOW() LIMIT ?;';
+// N.B., this sort order with limit is not actually fully deterministic
+const QUERY_EXPIRED_TOKENS =
+  'SELECT expiresAt, token, clientId FROM tokens WHERE expiresAt > ? AND expiresAt < NOW() ORDER BY expiresAt ASC, clientId, token LIMIT ?';
```
jrgm: Just to elaborate a bit on that cryptic comment, when I run the script against a million row table (generated with randomized expiresAt times), at then end, none of the protected clientId's are deleted and **almost** all of the target clientId rows **are** deleted.

But a few hundred remain.

The target rows that haven't been deleted are in tight bands of expiresAt time that are close to the end of each time window. I've dumped out all the rows that are found by each select and, yep, these "boundary" rows aren't found. But they should have been!. They have expiresAt that are older than rows that were found. I can't explain exactly why except to say 'indeterminate' (and some resources on the web do say that this type of range query with limit may exhibit this behaviour if using a non-unique index, expiresAt).

Anyways, I think this purgeExpiredTokensById method can be used to eliminate the bulk of the backlog and get the table size down to something less huge. And then the original version of purgeExpiredTokens can be used to pick up the stragglers.

jrgm: Or I have a minor bug in my code. But I don't see it. ;-)

rfk: Not sure it explains your symptons, but I feel like one of these `>` or `<` should be a `>=` or `<=`.

More generally, what happens if the `LIMIT 200` happens to cut off the result set halfway through a set of tokens that all expire at the same time?  E.g. if there are 300 tokens that all expire at the same time, does this delete the first 200 of them and then skip of the rest?

Also, we might want to expand the index to include the `token` and `clientId` columns so the sort can be satisfied completely from the index:

```
mysql> explain SELECT expiresAt, token, clientId FROM tokens WHERE expiresAt > '2018-00-00' AND expiresAt < NOW() ORDER BY expiresAt ASC, clientId, token LIMIT 200;
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
| id | select_type | table  | type | possible_keys | key  | key_len | ref  | rows | Extra                       |
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
|  1 | SIMPLE      | tokens | ALL  | idx_expiresAt | NULL | NULL    | NULL |   23 | Using where; Using filesort |
+----+-------------+--------+------+---------------+------+---------+------+------+-----------------------------+
1 row in set, 1 warning (0.01 sec)
```

jrgm:
Well, duh. The "tight bands" were 7 hours wide. ("tight" on the order of 2 weeks, I guess, but anyways). What else is 7 hours? My UTC offset.

I was being all smart and adjusting with `moment.utc`, but I shouldn't have really. So I've removed that and **all** expired tokens for non-pocket rows are removed, and **no** pocket rows are removed, expired or not. I've pushed up that change to this branch.

I also changed the queries to be inclusive of the range, since it felt right. It didn't change my results with my test runs, but might in production with more fine-grained expiresAt data. (Although, depending what `expiresAt` means, maybe those comparisons should stay as strictly less than?)

lib/db/mysql/index.js
jrgm: Yeah, could add that index. But if this job can run productively enough without it, I'd rather defer adding an index until I've reduced the row count (or not adding any index, if only this query would benefit. Actually, now that I think of it, the `client, token` part was only added as I flailed around trying to "fix" the ordering that was causing my "bug").

jrgm:
Pushed another fix to `s/ORDER BY expiresAt ASC, clientId, token/ORDER BY expiresAt/' which is 33% faster on a test data set.

jrgm:
... and has the same result.

lib/db/mysql/index.js
```diff
@@ -606,38 +614,159 @@ MysqlStore.prototype = {
...
+        // unknown ignoreClientId(s).
+        results.forEach((ignoreClient) => {
+          if (! ignoreClient) {
+            throw new Error('unknown ignoreClientId ' + ignoreClientId);
+          }
+        });
+      })
+      .then(() => {
+        // Continue from the last recorded 'last-purge-time', if available.
+        return self._write(QUERY_LAST_PURGE_TIME)
```
rfk: nit: this is a read, should we use `self._read` for this?  They currently do the same thing (just calling `this._query` but they may not in the future.

rfk: Actually `self._readOne` would be even better here and would avoid the need to defer `res[0]`.

test/db/purgeExpiredTokens.js
```diff
@@ -0,0 +1,238 @@
...
+          // Check the total tokens
+          return db._read('SELECT COUNT(*) AS count FROM fxa_oauth.tokens;');
+        })
+        .then( function (result) {
+          assert.equal(result[0].count, 1500);
+        });
+    });
+
+    if (purgeMethod === db.purgeExpiredTokens) {
+      // purgepurgeExpiredTokensById cannot meet the expectations of this
```
rfk: Typo "purgepurge"?

lib/db/mysql/index.js
jrgm: Updated to use readOne, or read where I want a list. Thanks.

test/db/purgeExpiredTokens.js
jrgm: Fixed this copy pasta.

PR #591, opened 2018-08-02
[WIP] feat(pairing): Make the FxA settings client "public" so that keys can be requested.

shane-tomlinson:
blocks https://github.com/mozilla/fxa-content-server/pull/6404

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -79,7 +79,10 @@
       "imageUri": "https://example2.domain/logo",
       "redirectUri": "https://example2.domain/return?foo=bar",
       "trusted": true,
-      "canGrant": true
+      "canGrant": true,
+      "trusted": true,
+      "publicClient": true,
+      "allowedScopes": "https://identity.mozilla.com/apps/oldsync"
     },
     {
       "name": "FxA OAuth Console",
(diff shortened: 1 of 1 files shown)


rfk:
FWIW, I think this change makes sense, and I'd also support adding some special logic so that `canGrant: true` clients are also allowed all scopes automatically, rather than having to list them in `allowedScopes`.

vladikoff:
@rfk to change this PR...

rfk:
@shane-tomlinson what's the difference between the "Firefox Accounts" and "Firefox Accounts Settings" reliers here?

On reflection, I don't think we should make this change - fxa-content-server shouldn't need to do anything *on behalf of itself* that requires `publicClient: true`.  Instead, it should be acting on behalf of whatever OAuth client is using the pairing flow, and should send that `client_id` in its requests to the oauth-server.

For initial dev and testing purposes, we can use an existing `client_id` that has the right configuration, such as `22d74070a481bc73` for `"Test Client iOS"`, which is already marked as public and able to request the `oldsync` scope.

shane-tomlinson:
This PR was meant to move the rest of things forward and not for merge to master, I imagine a new client_id for Sync is needed.

> @shane-tomlinson what's the difference between the "Firefox Accounts" and "Firefox Accounts Settings" reliers here?

I have no idea, @vladikoff, any ideas?

rfk:
I'm putting this in the backlog since we don't need it for the pairsona milestone.

vladikoff:
Closing for now, per moving repos

PR #599, opened 2018-09-05
Pairsona feature branch

rfk:
This is https://github.com/mozilla/fxa-oauth-server/pull/591/files re-opened as a feature branch while I work on getting it stood up at https://pairsona.dev.lcip.org


shane-tomlinson:
@rfk - this can be killed off now that we have a separate app for it, right?

rfk:
yeap

Issue #597, opened 2018-08-22
Add `npm audit` support


shane-tomlinson:
Add to travis, do not fail at first.

Ref #596

PR #603, opened 2018-09-18
chore(deps): Update dependencies to resolve `npm audit` warnings.

rfk:
Removing `grunt-nsp` and running `npm audit fix --force` appears to resolve all `npm audit` warnings while still allowing all the tests to pass.  I'm happy to defer this to a follow-up if we're worried about any of the updates, but otherwise seems we may as well do this as part of #602.

(Connects to #597)


rfk:
> appears to resolve all npm audit warnings while still allowing all the tests to pass

OK, well, CI apparently disagrees!

rfk:
Closing this since it works on my machine but not in CI, and I don't intend to spend any more time on it in the short term

PR #607, opened 2018-09-25
feat(openid): add `fxa-profileChangedAt` to /verify

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2490

This PR adds the `fxa-profileChangedAt` claim to the id token, and stores this value in the codes, refreshToken, tokens table. While the data is a bit duplicated, it is consistent with how things work on the oauth server.

The profile server will use the `profileChangedAt` value from tokens and compare it to the value return by the auth-server `/account/profile` route. If it is older then it will invalidate the profile server cache and fetch latest.


vbudhram:
@mozilla/fxa-devs The approach taken here is to stash the value in the codes table and use it when generating the id token. Does this make sense? or is there another way?

lib/db/mysql/patches/patch-021-022.sql
```diff
@@ -0,0 +1,4 @@
+-- Add column to stash the `profileChangedAt` value
+ALTER TABLE codes ADD COLUMN profileChangedAt BIGINT DEFAULT 0;
+
+UPDATE dbMetadata SET value = '22' WHERE name = 'schema-patch-level';
```
shane-tomlinson: Is adding a column to the profile server's dbMetadata table a slow process?

vbudhram: I don't believe so, but I don't think it will hurt to add `ALGORITHM = INPLACE, LOCK = NONE;` here.

vbudhram: Oh, misread thought this was about the alter table above. Updating the `dbMetadata` is a pretty quick operation. FWIW, I still added `ALGORITHM = INPLACE, LOCK = NONE;` to alter.

vbudhram:
After working through profile-server, moving this back to WIP. ISTM, that I need to update the `/verify` route to return `profileChangedAt` to profile-server.

lib/db/mysql/patches/patch-021-022.sql
```diff
@@ -0,0 +1,8 @@
+-- Add column to stash the `profileChangedAt` value
+ALTER TABLE codes ADD COLUMN profileChangedAt BIGINT DEFAULT 0,
+ALGORITHM = INPLACE, LOCK = NONE;
+
+ALTER TABLE tokens ADD COLUMN profileChangedAt BIGINT DEFAULT 0,
+ALGORITHM = INPLACE, LOCK = NONE;
```
rfk: We may also want to add it to the `refreshTokens` table, because a refreshToken can be used to create a new accessToken, and we'll want to new accessToken to have a sensible value for `profileChangedAt`.

lib/db/mysql/patches/patch-021-022.sql
```diff
@@ -0,0 +1,8 @@
+-- Add column to stash the `profileChangedAt` value
+ALTER TABLE codes ADD COLUMN profileChangedAt BIGINT DEFAULT 0,
+ALGORITHM = INPLACE, LOCK = NONE;
```
rfk: Please also update `lib/db/mysql/schema.sql` to reflect these changes.

vbudhram:
@mozilla/fxa-devs I think this is ready for an official review!

lib/db/mysql/schema.sql
```diff
@@ -31,7 +31,8 @@ CREATE TABLE IF NOT EXISTS codes (
   scope VARCHAR(256) NOT NULL,
   createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   authAt BIGINT DEFAULT 0,
-  offline BOOLEAN DEFAULT FALSE
+  offline BOOLEAN DEFAULT FALSE,
+  profileChangedAt BIGINT DEFAULT 0
```
rfk: I wonder whether `0` or `NULL` would be a better default here.  It probably won't matter either way, but defaulting to `NULL` in the db and them omitting `profile_changed_at` for tokens where we don't know it may be a little more explicit.

vbudhram:
@rfk Thanks for review! Updated this to not store in token claims or return value in `/authorization` `/token` routes.

PR #612, opened 2018-10-08
fix(db): Drop foreign key constraints.

### LINKED ISSUE #611: Drop foreign key constraints
We have some `on delete cascade` foreign key constraints, but they're not providing much value and they're interfering with the operation of the gh-ost db migration tool.  Per [1] and friends, we're going to drop them in prod in order to let the latest migration apply successfully.  Let's also do that as an explicit migration here in the codebase to ensure consistency with what's in prod.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1495860#c9

rfk:
These constraints make DB migrations harder and they're really not providing us much value in practice.

Fixes #611.

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -163,7 +163,15 @@ const QUERY_CLIENT_UPDATE = 'UPDATE clients SET ' +
   'trusted=COALESCE(?, trusted), allowedScopes=COALESCE(?, allowedScopes), ' +
   'canGrant=COALESCE(?, canGrant) ' +
   'WHERE id=?';
-const QUERY_CLIENT_DELETE = 'DELETE FROM clients WHERE id=?';
+// This query deletes everythin related to the client, and is thus quite expensive!
+// Don't worry, it's not exposed to any production-facing routes.
+const QUERY_CLIENT_DELETE = 'DELETE clients, codes, tokens, refreshTokens, clientDevelopers ' +
+  'FROM clients ' +
+  'LEFT JOIN codes ON clients.id = codes.clientId ' +
+  'LEFT JOIN tokens ON clients.id = tokens.clientId ' +
+  'LEFT JOIN refreshTokens ON clients.id = refreshTokens.clientId ' +
+  'LEFT JOIN clientDevelopers ON clients.id = clientDevelopers.clientId ' +
+  'WHERE clients.id=?';
 const QUERY_CODE_INSERT =
   'INSERT INTO codes (clientId, userId, email, scope, authAt, amr, aal, offline, code, codeChallengeMethod, codeChallenge, keysJwe, profileChangedAt) ' +
   'VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)';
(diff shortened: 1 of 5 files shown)


lib/db/mysql/schema.sql
```diff
@@ -53,25 +51,22 @@ CREATE TABLE IF NOT EXISTS tokens (
 
 CREATE TABLE IF NOT EXISTS developers (
   developerId BINARY(16) NOT NULL,
-  FOREIGN KEY (developerId) REFERENCES developers(developerId) ON DELETE CASCADE,
-  clientId BINARY(8) NOT NULL,
-  FOREIGN KEY (clientId) REFERENCES clients(id) ON DELETE CASCADE,
-  createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+  email VARCHAR(255) NOT NULL,
+  createdAt TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  UNIQUE(email)
```
rfk: The schema here was wrong, it did not match what had actually been created by the historical db migrations.

lib/db/mysql/index.js
```diff
@@ -163,7 +163,14 @@ const QUERY_CLIENT_UPDATE = 'UPDATE clients SET ' +
...
   'WHERE id=?';
-const QUERY_CLIENT_DELETE = 'DELETE FROM clients WHERE id=?';
+const QUERY_CLIENT_DELETE = 'DELETE clients, codes, tokens, refreshTokens, clientDevelopers, developers ' +
+  'FROM clients ' +
+  'JOIN codes ON clients.id = codes.clientId ' +
+  'JOIN tokens ON clients.id = tokens.clientId ' +
+  'JOIN refreshTokens ON clients.id = refreshTokens.clientId ' +
+  'JOIN clientDevelopers ON clients.id = clientDevelopers.clientId ' +
+  'JOIN developers ON clients.id = developers.clientId ' +
+  'WHERE clients.id=?';
```
jrgm: Is it worth noting in a comment that this query is on a route that is only available in dev mode?

lib/db/mysql/index.js
```diff
@@ -188,7 +195,10 @@ const QUERY_REFRESH_TOKEN_DELETE_USER =
   'DELETE FROM refreshTokens WHERE userId=?';
 const QUERY_CODE_DELETE_USER = 'DELETE FROM codes WHERE userId=?';
 const QUERY_DEVELOPER = 'SELECT * FROM developers WHERE email=?';
-const QUERY_DEVELOPER_DELETE = 'DELETE FROM developers WHERE email=?';
+const QUERY_DEVELOPER_DELETE = 'DELETE developers, clientDevelopers ' +
+  'FROM developers ' +
+  'JOIN clientDevelopers ON developers.developerId = clientDevelopers.developerID ' +
+  'WHERE developers.email=?';
```
jrgm: Actually, it looks like, aside from tests, nothing exposes `removeDeveloper` to a route (or used in script).

jrgm:
Also, as I said in IRC, we could leave the FK on {developers,clientDevelopers}, as those tables are really small.

rfk:
> Also, as I said in IRC, we could leave the FK on {developers,clientDevelopers}, as those tables are really small.

I'm going to remove them all for completeness, and to avoid future confusion.

rfk:
OK, I think this is ready - @vbudhram r? for another set of eyes?

lib/db/mysql/patches/patch-022-023.sql
```diff
@@ -0,0 +1,55 @@
...
+--  1) Explicitly create the indexes we need.  This "might" cause the ones
+--     that were created automatically for the FK constraint to be dropped.
+--
+--  2) Drop the FK constraints, which might leave behind the auto-created
+--     indexes if they weren't dropped in (1) above.
+--
+-- In my testing, the auto-created indexes are indeed dropped in favour
+-- of the explicit ones.  If they aren't, then at least we wind up with
+-- duplicate indexes which can be cleaned up manually, which is much better
+-- than winding up with no indexes at all.
```
vbudhram: 👍

lib/db/mysql/index.js
```diff
@@ -188,7 +195,10 @@ const QUERY_REFRESH_TOKEN_DELETE_USER =
   'DELETE FROM refreshTokens WHERE userId=?';
 const QUERY_CODE_DELETE_USER = 'DELETE FROM codes WHERE userId=?';
 const QUERY_DEVELOPER = 'SELECT * FROM developers WHERE email=?';
-const QUERY_DEVELOPER_DELETE = 'DELETE FROM developers WHERE email=?';
+const QUERY_DEVELOPER_DELETE = 'DELETE developers, clientDevelopers ' +
+  'FROM developers ' +
+  'JOIN clientDevelopers ON developers.developerId = clientDevelopers.developerID ' +
```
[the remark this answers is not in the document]
rfk: Yep, good catch, thanks 👍

PR #617, opened 2018-10-17
feat(server): update to Hapi 17

vladikoff:


GitCop:

 * Commit: b287fc840aff7b6c555ca3140765a67480a43d1b

---

lib/server/internal.js
```diff
@@ -12,17 +12,14 @@ const logger = require('../logging')('server.clients');
...
 const summary = require('../logging/summary');
 
-exports.create = function createServer() {
+exports.create = async function createServer() {
   var isProd = env.isProdLike();
 
-  var server = new Hapi.Server(
-    require('./config')
-  );
+  var serverConfig = require('./config');
```
vbudhram: nit: `const`

lib/routes/root.js
```diff
@@ -31,16 +31,24 @@ module.exports = {
...
-      var configPath = path.join(gitDir, 'config');
-      var cmd = 'git config --get remote.origin.url';
-      exec(cmd, { env: { GIT_CONFIG: configPath, PATH: process.env.PATH }}, function(err, stdout) {
-        source = stdout.replace(/\s+/, '');
-        return sendReply();
+    function runGitCmd() {
+      return new Promise((resolve) => {
+        // figure it out from .git
+        var gitDir = path.resolve(__dirname, '..', '..', '.git');
+        exec('git rev-parse HEAD', { cwd: gitDir }, function(err, stdout) {
```
vbudhram: Should this reject if an error is found?

vbudhram: Might be able to clean up some of callback nesting with http://bluebirdjs.com/docs/api/promise.fromcallback.html, but meh.

bin/internal.js
```diff
@@ -9,8 +9,10 @@ const server = require('../lib/server/internal').create();
 
 logger.debug('config', config);
 db.ping().done(function() {
-  server.start(function() {
-    logger.info('listening', server.info.uri);
+  server.then((s) => {
+    s.start(function() {
```
vbudhram: If we are changing lines, might be worthwhile to update to newer syntax?

vladikoff: 👍

lib/routes/root.js
vladikoff: I think in this case we just don't report the information for the `source` and etc

PR #619, opened 2018-10-24
Monorepo into auth server

rfk:
This prepares the fxa-oauth-server repo for being imported into fxa-auth-server; ref https://github.com/mozilla/fxa-auth-server/issues/2679


rfk:
The auth-server merge has landed, this should be good to go now.

Issue #620, opened 2018-10-25
Prepare a train-123 point-release that removes tokens.profileChangedAt


rfk:
As @jrgm noted in email, the migration from https://github.com/mozilla/fxa-oauth-server/pull/607 is running into some difficulty in production due to the enormous size of the `tokens` table.  We may need to revert that part of the change in order to unblock the rest of train-123.

In preparation for that eventuality, let's:

* Sanity-check that the related changes in https://github.com/mozilla/fxa-auth-server/pull/2638 and https://github.com/mozilla/fxa-profile-server/pull/343 will not break if we remove `profileChangedAt` support from oauth-server.  I *think* we coded them defensively so that the servers could be rolled out in any order, but it's worth checking.
* Prep a PR against train-123 branch which:
  * Drops the `profileChangedAt` column from the `tokens` table, as a new migration
  * Updates the db code so that `fxa-profileChangedAt` data from the assertion is silently dropped on the floor rather than trying to write it into a column that doesn't exist

If it turns out that we can't apply the migration in prod, we can merge the above PR in order to sync up the database state between prod and our dev environments, then deploy the point-release against the production database to unblock train-123.

We can then do another migration to bring back the `profileChangedAt` column in a future train, once we've successfully purged expired data from the `tokens` table in production.

@jrgm does this capture what we discussed earlier?

@vbudhram thoughts?

rfk:
> mozilla/fxa-auth-server#2638

I'm pretty confident that rolling back the change in oauth-server won't break anything in auth-server, since auth-server only records this information and pushes it out in the identity certificate, it doesn't expect to be able to receive any profile-changed-at information back from other services.

>  mozilla/fxa-profile-server#343 

If we back out the profile-changed-at functionality in oauth-server, then in the code here:

  https://github.com/mozilla/fxa-profile-server/pull/343/files#diff-741b62ef36092821244dbc9e55bc1af4R82

We will have `creds.profileChangedAt` as `undefined` while `result.profileChangedAt` will be a real timestamp.  In javascript, nothing will evaluate less than `undefined`, e.g.:

```
> 12345 < undefined
false
> Date.now() < undefined
false
```

Thus the `(result.profileChangedAt < creds.profileChangedAt)` condition will never be true, and we will not attempt to purge the profile cache.

So I think the auth and profile changes should be safe in the face of rollback.  We can test that out explicitly in stage if we need to go this route.

jrgm:
[...]

PR #621, opened 2018-10-25
fix(profile): remove the `profileChangedAt` colum on tokens table

vbudhram:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/620

This PR removes the `profileChangedAt` column on the tokens table. Is this what you had in mind @rfk?

### DIFF
--- a/lib/db/mysql/patches/patch-023-024.sql
+++ b/lib/db/mysql/patches/patch-023-024.sql
@@ -1,0 +1,8 @@
+-- This removes the `profileChangedAt` column on the tokens table.
+-- Since the tokens table is so large, this migration causes some issues.
+-- When the tokens get purged and the table is a bit smaller we can attempt to
+-- add this column.
+ALTER TABLE tokens DROP COLUMN profileChangedAt,
+ALGORITHM = INPLACE, LOCK = NONE;
+
+UPDATE dbMetadata SET value = '24' WHERE name = 'schema-patch-level';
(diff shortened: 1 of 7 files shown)


rfk:
Yep, thanks @vbudhram!  I haven't reviewed in detail but this looks like what we need.  /cc @jrgm for consideration.

jrgm:
Yes, looks right to me, from what I understand.

rfk:
OK, let's keep this in our pocket and see how the attempt at the migration goes come tomorrow...

lib/db/mysql/patches/patch-023-024.sql
```diff
@@ -0,0 +1,8 @@
+-- This removes the `profileChangedAt` column on the tokens table.
+-- Since the tokens table is so large, this migration causes some issues.
+-- When the tokens get purge and the table is a bit smaller we can attempt to
```
rfk: sp: "get purge" -> "get purged"

test/api.js
```diff
@@ -2886,7 +2886,7 @@ describe('/v1', function() {
           assert.equal(res.result.client_id, clientId);
           assert.equal(res.result.scope[0], 'profile');
           assert.equal(res.result.email, undefined);
-          assert.equal(res.result.profile_changed_at, PROFILE_CHANGED_AT_LATER_TIME, 'profile changed at is correct');
+          assert.equal(res.result.profile_changed_at, undefined, 'profile changed at is not set');
```
rfk, line 2893: I wonder if it's worth a comment on this test to explain why it's not set, in case we don't end up getting to this work for a little while.

rfk (approved):
I left a couple of nits but this looks solid, thanks @vbudhram!

> fix(profile): remove the `profileChangedAt` colum on tokens table

Typo in commit message "colum" -> "column".

PR #623, opened 2018-10-29
fix(db): Restore foreign key constraints on core tables.

rfk:
The migration removing these constraints did not stick in production, so we're reverting it here in code to keep things consistent.

Connects to https://github.com/mozilla/fxa-auth-server/issues/2695

### DIFF
--- a/lib/db/mysql/patch.js
+++ b/lib/db/mysql/patch.js
@@ -6,4 +6,4 @@
 // Update this if you add a new patch, and don't forget to update
 // the documentation for the current schema in ../schema.sql.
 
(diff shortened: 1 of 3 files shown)


rfk:
@jrgm does this match what was done in production?

jrgm:
> @jrgm does this match what was done in production?

Yes, that was what done in production for the `tokens` and `codes` tables, but not for `refreshTokens`. I'll restore that FK on refreshTokens on (my) Monday.

jrgm (approved):
r+

